PR #9, opened 2015-11-11
Support VTC in ad managers

angieisrani:
@pilievOoyala

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -179,9 +179,6 @@ OO.Ads.manager((function(_, $) {
         freeWheelCompanionAdsWrapper = $('#' + freeWheelCompanionAdsWrapperId);
       }
 
-      // TODO: Consider binding and unbinding this each time an ad is played
-      amc.ui.adVideoElement.bind('timeupdate', _.bind(fw_videoTimeUpdate, this));
-
       _positionOverlays();
     };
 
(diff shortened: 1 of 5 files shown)


js/freewheel.js
```diff
@@ -464,28 +461,6 @@ OO.Ads.manager((function(_, $) {
...
-     * @public
-     * @method Freewheel#resumeAd
-     * @param {object} ad The ad to resume
-     */
-    this.resumeAd = function(ad) {
-      OO.log("FW: Resuming the ad");
-      amc.ui.adVideoElement[0].play();
-    };
-
-    /**
```
aeng7: I noticed that the pause and resume functions are removed from Freewheel, but remain in Vast with an empty implementation and a comment saying VTC will handle them. Should we keep the functions in Freewheel and have a comment saying why there is no implementation? Just to get more information visible. Minor thing.

aeng7:
LGTM

js/freewheel.js
angieisrani: Yeah, I'm conflicted about this.  You're right that they should be consistent.  I had them in there but I thought it might be confusing since I wasn't using them.  I'm thinking I should remove them from VAST as well - what do you think?

aeng7: I'm actually leaning more towards having them both in. Since these will be public, someone might notice why they are missing and won't have an immediate answer.  Or we could not include them but have a comment (in code or documentation) saying why pause and resume are not implemented by specific ad managers.

aeng7:
LGTM

PR #12, opened 2015-11-24
Pbi 1216

aeng7:
-ima-vtc integration for non-shared stream elements (desktop webkit, Android)
-shared stream implementation will come in the near future

Notes/questions:
-what should default volume and currentTime be if required components are not setup yet (for getVolume and getCurrentTime)?

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -77,6 +77,8 @@ OO.Ads.manager(function(_, $)
     {
       _amc = amcIn;
 
+      OO.Video.plugin(new GoogleIMAVideoFactory());
+
       var ext = _amc.platform.DEV ? '_debug.js' : '.js';
       remoteModuleJs = "//imasdk.googleapis.com/js/sdkloader/ima3" + ext;
       _resetVars();
@@ -121,6 +123,9 @@ OO.Ads.manager(function(_, $)
       _linearAdIsPlaying = false;
       _resetPlayheadTracker();
       this.hasPreroll = false;
+
+      this.adPlaybackStarted = false;
+      this.vcPlayRequested = false;
 
       //flag to track whether ad rules failed to load
       this.adRulesLoadError = false;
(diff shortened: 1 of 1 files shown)


angieisrani:
Good questions:  Assume volume=1.0 (100%) if not configured.  Assume currentTime=0 if not played yet unless setInitialTime called.

js/google_ima.js
```diff
@@ -1414,13 +1518,241 @@ OO.Ads.manager(function(_, $)
...
+    this.maxSupportedElements = -1;
+
+    /**
+     * Returns the number of video elements currently instantiated.
+     * @public
+     * @method GoogleIMAVideoFactory#getCurrentNumberOfInstances
+     * @returns {int} The number of video elements created by this factory that have not been destroyed
+     */
+    this.getCurrentNumberOfInstances = function() {
+    };
```
angieisrani: This function is being removed - you can take it out of here

angieisrani:
So far looks good to me.  Mind walking through a couple pieces with me in person?

aeng7:
Sure, will do.

aeng7:
Support for multiple players has been implemented.

aeng7:
Since I moved the old Ad Manager code inside a function, the diff is hard to parse. The main changes:
-changed default values for getCurrentTime and getVolume inside GoogleIMA
-added a playerId parameter to the GoogleIMA object's initialize method, which is used to store the instance inside registeredGoogleIMAManagers
-registeredGoogleIMAManagers will be used to retrieve instances when the Factory's create method is called. The create method now includes the playerId parameter

These changes were made to support multiple players on a page.

aeng7:
You can also do a local diff and pull the OO.Ads.manager(); block of code outside of and above the outer anonymous function. The diff should be much more readable that way.

angieisrani:
Looks great to me - merging

angieisrani:
@aeng7 I missed one thing - can you update the template with the new parameter for the initialize function and open a new pull request?

aeng7:
Will do.

PR #13, opened 2015-11-25
[PBI-1252] Fix issue where preroll overlay was being time out by AMC

dchauOoyala:
- Issue was caused by a TypeError in the fw_onSlotStarted function
- Similar issue found in fw_onSlotEnded function
- Add check to prevent TypeError in the event that adVideoElement is null or if the ad is non-linear

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -402,6 +402,7 @@ OO.Ads.manager(function(_, $) {
           // Trigger the ad
           _registerDisplayForNonlinearAd();
           _registerDisplayForLinearAd();
+          debugger;
           currentPlayingSlot = ad.ad;
           indexInPod = 0;
           if (ad.isLinear) {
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -385,6 +386,7 @@ OO.Ads.manager(function(_, $) {
      */
     this.playAd = function(ad) {
       _resetAdState();
+      currentAd = ad;
```
angieisrani: We already save currentPlayingSlot - are you able to use that?

dchauOoyala: I did see the currentPlayingSlot variable. That get's assigned as ad.ad on line 407, which no longer has the isLinear field. Is there something similar to isLinear from ad.ad?

angieisrani: Slot has getTimePositionClass() which can tell us if it's a preroll, midroll, postroll, or overlay.  Technically it does not differentiate between linear and nonlinear overlays so we assume the overlay to be nonlinear

dchauOoyala: All right, I'll change it and use getTimePositionClass() instead

js/freewheel.js
```diff
@@ -742,7 +742,12 @@ OO.Ads.manager(function(_, $) {
      * @method Freewheel#fw_onSlotStarted
      */
     var fw_onSlotStarted = function() {
-      amc.ui.adVideoElement.removeAttr('controls');
+      // adVideoElement may be null for overlays
+      if (currentPlayingSlot &&
+          currentPlayingSlot.getTimePositionClass() === tv.freewheel.SDK.TIME_POSITION_CLASS_OVERLAY &&
```
pilievOoyala: isn't this supposed to check if it's a linear ad (not overlay) and then remove the attribute?

dchauOoyala: you're right

gfrankOoyala:
Looks good to me.

PR #16, opened 2015-12-01
[PBI-1191] Call function to raise WILL_PLAY_NONLINEAR_AD event in Google IMA and Freewheel

dchauOoyala:


js/google_ima.js
```diff
@@ -425,7 +425,8 @@ require("../html5-common/js/utils/utils.js");
           //displaying it, so just notify AMC that we are showing an overlay.
           else
           {
-            _amc.notifyNonlinearAdStarted(this.currentAMCAdPod.id);
+            // raise WILL_PLAY_NONLINEAR_AD event and notify nonlinear ad started
+            _amc.sendURLToLoadAndPlayNonLinearAd(this.currentAMCAdPod, this.currentAMCAdPod.id, this.currentAMCAdPod.ad.tag_url);
```
angieisrani: Does this result in the ooyala skin rendering the nonlinear ad?

gfrankOoyala: If it doesn't do we need to send ad.tag_url? If not we should send null and update the comment to indicate it is alerting the AMC and player that a non linear ad is started but it is rendered by ima.

angieisrani: Also, do we no longer need to call notifyNonlinearAdStarted?  Isn't that what is triggering the WILL_PLAY_NONLINEAR_AD event?  And actually, if we're calling that, do we need to call sendURLToLoadAndPlayNonLinearAd at all or is sendURLToLoadAndPlayNonLinearAd replacing notifyNonlinearAdStarted?

gfrankOoyala: That is called by sendURL, I just double checked.

dchauOoyala: In regards to the url, I was not aware of its implications. I saw that in
vast, the streamurl was passed into the sendURL function. Can you guys tell
me what the difference is?

And yes, sendURL does have the notify fun toon within it so the sendURL
does replace it.

gfrankOoyala: If the url is there, then it will send that to the player, which it will use to render the overlay. In VAST it will render the image that i sent via the streamurl. So it is best to send null if it isn't going to render it by the player but by the sdk to avoid any confusion or mistakes.

dchauOoyala: I see, thanks for the explanation!

js/freewheel.js
```diff
@@ -429,10 +429,10 @@ OO.Ads.manager(function(_, $) {
           } else {
             fwContext.setParameter("renderer.video.clickDetection", true, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
             slotStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(adId) {
-                amc.notifyNonlinearAdStarted(adId);
+                amc.sendURLToLoadAndPlayNonLinearAd(ad, ad.id, null);
               }, this, ad.id);
             slotEndedCallbacks[ad.ad.getCustomId()] = _.bind(function(adId) {
-                amc.notifyNonlinearAdEnded(adId);
+                amc.sendURLToLoadAndPlayNonLinearAd(ad, ad.id, null);
```
gfrankOoyala: You don't want to change that here :), thisis that the linear ad ended, diferent callback, please revert this line

js/google_ima.js
```diff
@@ -425,7 +425,9 @@ require("../html5-common/js/utils/utils.js");
           //displaying it, so just notify AMC that we are showing an overlay.
           else
           {
-            _amc.notifyNonlinearAdStarted(this.currentAMCAdPod.id);
+            // raise WILL_PLAY_NONLINEAR_AD event and alert AMC and player that a nonlinear ad is started.
+            // Nonlinear ad is rendered by IMA.
+            _amc.sendURLToLoadAndPlayNonLinearAd(this.currentAMCAdPod, this.currentAMCAdPod.id, null);
```
gfrankOoyala: looks good.

js/freewheel.js
dchauOoyala: facepalm... will fix.

gfrankOoyala:
LGTM

PR #18, opened 2015-12-02
pbi-1283

aeng7:
-ad marquee and scrubber will now properly reflect the current time for IMA ads
-moved videoControllerWrapper function calls into main event handler switch to better control the flow of events. This provides for a better user experience and less issues down the road. One example is previously we were going through the entire endCurrentAd flow (including publishing the relevant ad end messages) before we would notify the video controller that the ad was ending

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -557,7 +557,10 @@ require("../html5-common/js/utils/utils.js");
         //IMA provides values for getRemainingTime which can result in negative current times
         //or current times which are greater than duration.
         //We will check these boundaries so we will not report these unexpected current times
-        if (_IMAAdsManager && this.currentIMAAd && _IMAAdsManager.getRemainingTime() >= 0 && _IMAAdsManager.getRemainingTime() <= this.currentIMAAd.getDuration())
+        if (_IMAAdsManager &&
+          this.currentIMAAd &&
+          _IMAAdsManager.getRemainingTime() >= 0 &&
+          _IMAAdsManager.getRemainingTime() <= this.currentIMAAd.getDuration())
         {
           currentTime = this.currentIMAAd.getDuration() - _IMAAdsManager.getRemainingTime();
         }
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -90,7 +90,7 @@ require("../html5-common/js/utils/utils.js");
         _amc = amcIn;
 
         var ext = _amc.platform.DEV ? '_debug.js' : '.js';
-        remoteModuleJs = "//imasdk.googleapis.com/js/sdkloader/ima3" + ext;
+        var remoteModuleJs = "//imasdk.googleapis.com/js/sdkloader/ima3" + ext;
         _resetVars();
         _createAMCListeners();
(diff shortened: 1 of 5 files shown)


js/google_ima.js
```diff
@@ -1191,6 +1229,12 @@ require("../html5-common/js/utils/utils.js");
               }, this), 100);
             }
             break;
+          case eventType.PAUSED:
+            if (this.videoControllerWrapper)
```
gfrankOoyala: does the timerupdate need to be paused?

js/google_ima.js
```diff
@@ -1212,35 +1256,55 @@ require("../html5-common/js/utils/utils.js");
...
+      /**
+       * Starts a timer that will provide an update to the video controller wrapper of the ad's current time.
+       * We use a timer because IMA does not provide us with a time update event.
+       * @private
+       * @method GoogleIMA#_startTimeUpdater
+       */
+      var _startTimeUpdater = privateMember(function()
+      {
+        _stopTimeUpdater();
+        _timeUpdater = setInterval(_.bind(function()
```
gfrankOoyala: Are we able to access the video object of the ad and add a playhead time changed or is that guarded by the SDK?

js/google_ima.js
```diff
@@ -548,13 +551,26 @@ require("../html5-common/js/utils/utils.js");
       this.getCurrentTime = function()
       {
         var currentTime = 0;
-        if (_IMAAdsManager && this.currentIMAAd)
+        //IMA provides values for getRemainingTime which can result in negative current times
+        //or current times which are greater than duration.
+        //We will check these boundaries so we will not report these unexpected current times
+        if (_IMAAdsManager && this.currentIMAAd && _IMAAdsManager.getRemainingTime() >= 0 && _IMAAdsManager.getRemainingTime() <= this.currentIMAAd.getDuration())
```
angieisrani: may want to link break this based on style guides

js/google_ima.js
angieisrani: From the main_html5 plugin side I can say that the events are not raised while paused.
I think the only harm they will have is possibly raising events that aren't needed, which may be a big issue or a small one depending on performance impact.  I'm guessing it will be fine either way

angieisrani:
overall looks good to me!

js/google_ima.js
aeng7: I don't pause the timerupdate when we get a PAUSED event in case the playhead changes while paused (such as for scrubbing). This ensures that while the video is on screen (between STARTED and COMPLETED), the playhead time is correct.

js/google_ima.js
aeng7: Unfortunately we have no way of retrieving the video object. :(

js/google_ima.js
aeng7: Done!

js/google_ima.js
gfrankOoyala: We don't scrub ads, but I see what you are saying, that is fine.

gfrankOoyala:
LGTM

PR #19, opened 2015-12-05
[PBI-1264] Updating streams to be sent to VTC based on what is available

gfrankOoyala:


js/ad_manager_vast.js
```diff
@@ -369,16 +369,24 @@ OO.Ads.manager(function(_, $) {
...
       var linearAd = ad.data.linear;
       if (linearAd && linearAd.mediaFiles) {
         var vastStreams = linearAd.mediaFiles;
-        streams["webm"] = this._extractStreamForType(linearAd.mediaFiles, "webm");
-        if (ad.streamUrl == null) {
-          streams["mp4"] = this._extractStreamForType(linearAd.mediaFiles, "mp4");
-          OO.log("extract m3u8 stream here"); // TODO
+        var videoEncodingsSupported = OO.VIDEO.ENCODING;
+        for (var encoding in videoEncodingsSupported) {
+          streamData = null;
```
angieisrani: You could consider moving the variable declaration into this scope but it's not important

js/ad_manager_vast.js
```diff
@@ -369,16 +369,24 @@ OO.Ads.manager(function(_, $) {
...
+          if (videoEncodingsSupported.hasOwnProperty(encoding)) {
```
angieisrani: What situation does this protect from?  The encoding property is coming from the same variable.

js/ad_manager_vast.js
```diff
@@ -369,16 +369,24 @@ OO.Ads.manager(function(_, $) {
...
+            streamData = this._extractStreamForType(vastStreams, videoEncodingsSupported[encoding]);
+          }
+          if (streamData) {
+            streams[videoEncodingsSupported[encoding]] = streamData;
```
angieisrani: can you do this directly in the if above?

js/ad_manager_vast.js
```diff
@@ -369,16 +369,24 @@ OO.Ads.manager(function(_, $) {
...
+            if (ad.streamUrl == null || videoEncodingsSupported[encoding] == OO.VIDEO.ENCODING.MP4) {
+              ad.streamUrl = streams[videoEncodingsSupported[encoding]];
```
angieisrani: What is the ad streamUrl?  And why do we prefer mp4?

js/ad_manager_vast.js
```diff
@@ -559,17 +567,8 @@ OO.Ads.manager(function(_, $) {
...
-      switch (type) {
-        case "webm":
-          filter.push("video/webm");
-           break;
-        case "mp4":
-          filter.push("video/mp4");
-          if (OO.isIos) { filter.push("video/quicktime"); }
-          break;
-      }
+      filter.push("video/" +type);
```
angieisrani: nice and simple!

js/ad_manager_vast.js
```diff
@@ -589,24 +588,14 @@ OO.Ads.manager(function(_, $) {
...
-      if (this.vastAdUnit.streamUrl == null && OO.supportedVideoTypes.mp4) {
-        this.vastAdUnit.streamUrl = this._extractStreamForType(streams, "mp4");
-      }
-      if (this.vastAdUnit.streamUrl == null && OO.supportedVideoTypes.m3u8) {
-        OO.log("extract m3u8 stream here");
-      }
       _.extend(this.vastAdUnit.data, firstLinearAd);
       this.vastAdUnit.data.tracking = firstLinearAd.linear.tracking;
-
+      addToTimeline(this.vastAdUnit, adLoaded);
```
angieisrani: Should we add to timeline even if this.vastAdUnit.streamUrl == null?

js/ad_manager_vast.js
gfrankOoyala: Ad stream url might not be used or only for analytics, not sure, but before it was used to play the video and we were defaulting to mp4, So I didn't want to remove that since I don't know where it is used. was thinking of removing it later when we build the a analytics framework and find out if it is needed or not.

js/ad_manager_vast.js
gfrankOoyala: yes but I was trying not to embed too many ifs. Do you think it should be moved?

js/ad_manager_vast.js
gfrankOoyala: To be honest, no idea, I saw this in an example online of how to do a proper for each, I can remove the if.

js/ad_manager_vast.js
gfrankOoyala: was thinking of doing that, will move it.

js/ad_manager_vast.js
gfrankOoyala: hmm, good question, probably not, because it means that there is no video url .

js/ad_manager_vast.js
gfrankOoyala: The only thing I can think is if someone puts in a constant with null value but that would be a bigger issue.

js/ad_manager_vast.js
gfrankOoyala: actually I think we should as it might be a non-linear ad and this doesn't look for stream url of non-linear.

gfrankOoyala: Nm, that would get set before call addToTimline. On it.

js/ad_manager_vast.js
gfrankOoyala: the above if is removed so no longer needed

js/ad_manager_vast.js
angieisrani: good point

js/ad_manager_vast.js
angieisrani: Oh I thought you may be able to omit this if but if not then this is fine

js/ad_manager_vast.js
angieisrani: I see.  I think since the VTC is selecting the stream this value could end up incorrect since the user may not favor mp4 (mp4 is not the top of the list by default).  I would suggest either reading the stream url from the videoControllerVideoElementCreated event in the AMC and passing it along to the ad object or omitting it entirely.  Might rather not have data than have wrong data

gfrankOoyala: Hmm, not sure if this is needed for video urls or not. How do I get what VTC created? It would well after this function.

gfrankOoyala: Right now it seems like it is only used by the vast ad manager and for non-linear ads, so I am open to suggestions on what to do for video ads

js/ad_manager_vast.js
```diff
@@ -372,16 +367,24 @@ OO.Ads.manager(function(_, $) {
       var linearAd = ad.data.linear;
       if (linearAd && linearAd.mediaFiles) {
         var vastStreams = linearAd.mediaFiles;
-        streams["webm"] = this._extractStreamForType(linearAd.mediaFiles, "webm");
-        if (ad.streamUrl == null) {
-          streams["mp4"] = this._extractStreamForType(linearAd.mediaFiles, "mp4");
-          OO.log("extract m3u8 stream here"); // TODO
+        var videoEncodingsSupported = OO.VIDEO.ENCODING;
+        var streamData = null;
```
angieisrani: you may be able to define this within the for loop but it's not a big deal

js/ad_manager_vast.js
```diff
@@ -372,16 +367,24 @@ OO.Ads.manager(function(_, $) {
...
-      timeline.push(new this.amc.Ad({position:positionSeconds, duration:duration, adManager:this.name,
-                                     ad:ad, adType:type, streams: streams}));
-      this.amc.appendToTimeline(timeline);
+      if (ad.streamUrl != null || (type == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(streams))) {
+        timeline.push(new this.amc.Ad({
+          position: positionSeconds, duration: duration, adManager: this.name,
+          ad: ad, adType: type, streams: streams
+        }));
+        this.amc.appendToTimeline(timeline);
+        ad.streams = streams;
```
angieisrani: Why not do this earlier?

js/ad_manager_vast.js
```diff
@@ -589,24 +583,14 @@ OO.Ads.manager(function(_, $) {
...
-      }
-      if (this.vastAdUnit.streamUrl == null && OO.supportedVideoTypes.m3u8) {
-        OO.log("extract m3u8 stream here");
-      }
       _.extend(this.vastAdUnit.data, firstLinearAd);
       this.vastAdUnit.data.tracking = firstLinearAd.linear.tracking;
-
-      if (this.vastAdUnit.streamUrl == null) {
+      addToTimeline(this.vastAdUnit, adLoaded);
+      if (_.isEmpty(this.vastAdUnit.streams)) {
```
angieisrani: not sure i understand why we add to timeline if there are no streams available (since this must be a linear ad) but I don't see this causing an issue

angieisrani: Sounds like the return value will not affect whether or not we should append to the timeline.  The return value tells us whether or not there were streams in the ad object that was appended to the timeline.

js/ad_manager_vast.js
gfrankOoyala: Moved to an earlier spot.

js/ad_manager_vast.js
gfrankOoyala: I would rather leave it outside for clarity sake, if you are strongly in the opinion that it should be outside I will move it.

angieisrani: no problem

angieisrani:
LGTM

PR #23, opened 2015-12-11
PBI-1227

aeng7:
Freewheel overlay placement fix. Uses undocumented API methods.

Creating a PR for this change. It should be safe to put in, but if Freewheel gets back to us with a better method, we can revert this and put in the new one.

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -75,6 +75,7 @@ OO.Ads.manager(function(_, $) {
       amc.addPlayerListener(amc.EVENTS.CONTENT_COMPLETED, _.bind(onContentCompleted, this));
       amc.addPlayerListener(amc.EVENTS.CONTENT_AND_ADS_COMPLETED, _.bind(onAllCompleted, this));
       amc.addPlayerListener(amc.EVENTS.CONTENT_CHANGED, _.bind(onContentChanged, this));
+      amc.addPlayerListener(amc.EVENTS.SIZE_CHANGED, _.bind(onResize, this));
     };
 
     /**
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -540,6 +541,34 @@ OO.Ads.manager(function(_, $) {
...
+      //Freewheel only attempts to resize ads and not re-position ads.
+      //In the Freewheel SDK, overlay ads do not have a resize function and thus nothing happens.
+
+      //We want to force the renderer to reposition the overlay ads. The Freewheel SDK has functions
+      //that accomplish this, but are undocumented.
+      if(currentPlayingSlot){
+        //These two calls update Freewheel with the new video sizes. At this point
+        //Freewheel will attempt to resize any ads
+        fwContext.setContentVideoElement(amc.ui.ooyalaVideoElement[0]);
+        fwContext.registerVideoDisplayBase(amc.ui.adWrapper.attr("id"));
```
angieisrani: Are you able to reuse these functions here:
           _registerDisplayForNonlinearAd();
           _registerDisplayForLinearAd();

aeng7: I thought about that as well. Only those specific functions (setContentVideoElement and registerVideoDisplayBase) need to be called, so if _registerDisplayForNonlinearAd(); and _registerDisplayForLinearAd(); get updated to include more in the future, we would need to update the onResize function to only call fwContext.setContentVideoElement(amc.ui.ooyalaVideoElement[0]); and
 fwContext.registerVideoDisplayBase(amc.ui.adWrapper.attr("id")); I'm fine with either, but I believe this will cause less headaches in the future.

angieisrani: That makes sense - but if those have to change, won't this have to change?  Basically there are several places where we have to call these two functions (and sometimes just one or the other) and i filtered it down to two basic concepts.  One sets up freewheel to properly render linear ads and the other sets it up to properly render nonlinear ads.  I think we should use those because if it ever changes there it will probably need to change here.  Based on my checking i think the commands and parameters are currently the same

aeng7: I'll add a function called notifySizeChange and call registerDisplayFor<>Ad in that function. That way we don't lose track of what we functionality we want to accomplish at this point.

aeng7:
Freewheel has acknowledged this limitation and will be putting it as a feature request for a future release. We should go ahead with this workaround for now.

angieisrani:
Quick question, in the first commit you call both:
+        fwContext.setContentVideoElement(amc.ui.ooyalaVideoElement[0]);
+        fwContext.registerVideoDisplayBase(amc.ui.adWrapper.attr("id"));
But in the second commit you call one or the other using an if.  Is this working?

aeng7:
Yep, this works fine. I wanted to isolate the calls with what they are tied with. I should have made a comment about this. Sorry about that.

angieisrani:
LGTM

PR #25, opened 2015-12-16
pbi-1300

aeng7:
-fixed an issue where we were not properly handling volume values across different ad playbacks


angieisrani:
looks good to me - can we sync up offline about when the vtc is not raising the volume event?

PR #26, opened 2015-12-17
pbi-1307

aeng7:
-workaround of an issue where overlays are not initially shown. We now use our update overlay position logic when the main video element comes into focus

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -76,6 +76,7 @@ OO.Ads.manager(function(_, $) {
       amc.addPlayerListener(amc.EVENTS.CONTENT_AND_ADS_COMPLETED, _.bind(onAllCompleted, this));
       amc.addPlayerListener(amc.EVENTS.CONTENT_CHANGED, _.bind(onContentChanged, this));
       amc.addPlayerListener(amc.EVENTS.SIZE_CHANGED, _.bind(onResize, this));
+      amc.addPlayerListener(amc.EVENTS.MAIN_CONTENT_IN_FOCUS, _.bind(onMainContentInFocus, this));
     };
 
     /**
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -541,7 +541,7 @@ OO.Ads.manager(function(_, $) {
       // if (fwContext && _.isFunction(fwContext.dispose)) fwContext.dispose();
     };
 
-    var onResize = function() {
+    var updateOverlayPosition = function() {
       //Overlay placement issue - PBI-1227 as of 12/9/2015
       //The main issue with Freewheel is when notifying their SDK of video size changes,
       //Freewheel only attempts to resize ads and not re-position ads.
```
angieisrani: May need to rename notifySizeChange() as well?

js/freewheel.js
```diff
@@ -634,6 +638,22 @@ OO.Ads.manager(function(_, $) {
...
+    this.videoElementInFocus = function(videoId) {
+      // The Freewheel SDK does not like when the video elements passed in via _registerDisplayForLinearAd
+      // and _registerDisplayForNonlinearAd have display set to none. This causes overlays to not
+      // be positioned and sized correctly (PBI-1307). When we get notified that the main video content
+      // is in focus, we need update the overlay position.
+      // Note: Updating overlay position uses undocumented methods and may break at any time. Also, simply
+      // notifying the SDK via _registerDisplayForNonlinearAd does not appear to be sufficient at this time
+      if (videoId === OO.VIDEO.MAIN) {
+        updateOverlayPosition();
+      }
```
angieisrani: Hm - the one thing i'm feeling mixed about is the fact that now freewheel has to be aware of what videoIds are.  Also I would advise using an ad manager "event" for notifications rather than a function so that it's only called on the subscribed plugins and is called on all plugins

js/freewheel.js
aeng7: I can rename these functions if there is some confusion. The change I made here was to pull out the logic for updating the overlay position into its own function (updateOverlayPosition) and call that inside onResize. I also call updateOverlayPosition when we get notified when the main video element comes into focus.

angieisrani: I suppose coming into focus might be considered a resize, i can see that

js/freewheel.js
aeng7: Agreed, will make this change.

aeng7:
Freewheel will now consume the MAIN_CONTENT_IN_FOCUS event to reposition overlays when the main content comes into focus.

angieisrani:
LGTM!  Can this be merged even if AMF changes are not in?

angieisrani:
Merging as the AMF change is merged

PR #28, opened 2015-12-18
pbi-1300

aeng7:
-fixed an issue where saved volume was not consumed properly (additional fix for pbi-1300)
-fixed and added unit tests dealing with volume

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -507,11 +507,6 @@ require("../html5-common/js/utils/utils.js");
           }
           else
           {
-            if (this.savedVolume >= 0)
-            {
-              this.setVolume(this.savedVolume);
-              this.savedVolume = -1;
-            }
             _IMAAdsManager.start();
             this.adPlaybackStarted = true;
           }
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -1245,6 +1240,11 @@ require("../html5-common/js/utils/utils.js");
             if(ad.isLinear())
             {
               _linearAdIsPlaying = true;
+              if (this.savedVolume >= 0)
```
gfrankOoyala: Would this still work if you change the volume after the ad starts, with or without pausing it?

aeng7: Yep if the ad is started, _linearAdIsPlaying is true and the setVolume will call the IMA API rather than saving it.

test/unit-tests/ima_test.js
```diff
@@ -767,11 +767,43 @@ describe('ad_manager_ima', function()
...
+  {
+    initAndPlay(true, vci);
+    var TEST_VOLUME = 0.5;
+    videoWrapper.setVolume(TEST_VOLUME);
+    expect(ima.savedVolume).to.be(TEST_VOLUME);
+  });
+
+  it('VTC Integration: Saved volume is consumed when ad starts', function()
+  {
+    debugger;
```
pilievOoyala: please remove this debugger statement.

pilievOoyala:
Just the one debug statement otherwise it's good and you can merge it when you commit it.

aeng7:
Removed debugger statement.

PR #29, opened 2015-12-22
Updating IMA plugin with new fields

aeng7:
-Updated IMA-VTC plugin to include features and technology.
-IMA plugin will not validate correctly without these


aeng7:
@angieisrani , please let me know if a feature should be added or the technology should be different. Thanks!

angieisrani:
This is perfect, sorry to omit it.  Please also add the properties so the jsdocs (i may have forgotten to do this on the other plugins as well)

angieisrani:
Merging to fix IMA - we can do jsdocs in another pull

PR #32, opened 2016-01-06
[PBW-3961][Chrome][FW] - PreMidPost - Looping after it replays

brianfwong:
In this.onReplay() within ad_manager_controller.js we call
_triggerAdManagerCallback(this.EVENTS.REPLAY_REQUESTED); which in turn
invokes the event listener for this event within the specific ad
manager (in this case freewheel ad manager). The event listener for
EVENTS.REPLAY_REQUESTED was missing in freewheel.js. Adding the event
listener onRePlayRequested fixed the issue. Appears to also solve
PBI-1259

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -75,7 +75,7 @@ OO.Ads.manager(function(_, $) {
     this.initialize = function(amcIn) {
       amc = amcIn;
       amc.addPlayerListener(amc.EVENTS.INITIAL_PLAY_REQUESTED, _.bind(onPlayRequested, this));
-      amc.addPlayerListener(amc.EVENTS.REPLAY_REQUESTED, _.bind(onRePlayRequested, this));
+      amc.addPlayerListener(amc.EVENTS.REPLAY_REQUESTED, _.bind(onReplayRequested, this));
       amc.addPlayerListener(amc.EVENTS.PLAY_STARTED, _.bind(onPlay, this));
       amc.addPlayerListener(amc.EVENTS.PAUSE, _.bind(onPause, this));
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -75,6 +75,7 @@ OO.Ads.manager(function(_, $) {
     this.initialize = function(amcIn) {
       amc = amcIn;
       amc.addPlayerListener(amc.EVENTS.INITIAL_PLAY_REQUESTED, _.bind(onPlayRequested, this));
+      amc.addPlayerListener(amc.EVENTS.REPLAY_REQUESTED, _.bind(onRePlayRequested, this));
```
pilievOoyala: I'm going to be 'that guy'. Since 'replay' is one word can you please rename it to onReplayRequested. It'll be consistent with our other files.

js/freewheel.js
```diff
@@ -671,6 +672,14 @@ OO.Ads.manager(function(_, $) {
       }
     };
 
+    var onRePlayRequested = function() {
+      _resetAdState();
+      shouldRequestAds = true;
```
pilievOoyala: Can you consolidate the inside of this function (except the _resetAdState()) to another private function. It's the same as onPlayRequested(). Then have both of them call this util function.  Call it something like 'setupAdsWrapper'.  Thank you!

js/freewheel.js
brianfwong: Change to function name made and checked in.

js/freewheel.js
brianfwong: @pilievOoyala Change made - created a new function and had the two other functions call it.

pilievOoyala:
Looks great. Thank you!

PR #34, opened 2016-01-07
[PBW-3665] Support VTC+iOS+IMA

angieisrani:
This should be merged at the same time as the changes to html5-video-plugins (https://github.com/ooyala/html5-video-plugins/pull/55) and the core player (PR 464 open in mjolnir)

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1771,6 +1771,8 @@ require("../html5-common/js/utils/utils.js");
    * @property {object} controller A reference to the Ooyala Video Tech Controller
    * @property {boolean} disableNativeSeek When true, the plugin should supress or undo seeks that come from
    *                                       native video controls
+   * @property {boolean} readyForCss When true, css may be applied on the video element.  When false, css
+   *                                 should be stored for use later when this value is true.
    */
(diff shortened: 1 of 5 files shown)


js/google_ima.js
```diff
@@ -272,6 +274,11 @@ require("../html5-common/js/utils/utils.js");
       this.registerUi = function()
       {
         this.uiRegistered = true;
+        if (_amc.ui.useSingleVideoElement && !this.sharedVideoElement && _amc.ui.ooyalaVideoElement[0] &&
+            (_amc.ui.ooyalaVideoElement[0].className === "video")) {
```
aeng7: Where is this className defined? It would be better if it can be fetched or received from somewhere. We can avoid issues that arise if this class name needs to be changed in the future.

aeng7:
LGTM if it takes some effort for the class name. We can address that at another time.

js/google_ima.js
angieisrani: Oh this is actually a property that native to DOM elements.  We could also do a get on the property but it would be something like getProperty("className") and I don't know if that's better than reading it directly.  Can you explain what you mean by fetched?

PR #35, opened 2016-01-08
Pbi 1302

aeng7:
-single video element mode will now use plugins div rather than ad wrapper when creating an IMA AdDisplayContainer. This allows the IMA UI (Learn more and Skip button) to properly render on iPad
-added a mock function for unit tests so they work again

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -935,11 +935,12 @@ require("../html5-common/js/utils/utils.js");
             _IMAAdDisplayContainer.destroy();
           }
 
-          var uiWrapper = this.sharedVideoElement ? _amc.ui.pluginsElement[0] : _amc.ui.adWrapper[0];
-
           //iphone performance is terrible if we don't use the custom playback (i.e. filling in the second param for adDisplayContainer)
           //also doesn't not seem to work nicely with podded ads if you don't use it.
-          _IMAAdDisplayContainer = new google.ima.AdDisplayContainer(uiWrapper,
+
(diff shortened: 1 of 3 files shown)


aeng7:
Should be merged the same time as mjolnir v4 PR 467 and after https://github.com/ooyala/html5-skin/pull/284 is merged.

js/google_ima.js
```diff
@@ -937,7 +937,10 @@ require("../html5-common/js/utils/utils.js");
 
           //iphone performance is terrible if we don't use the custom playback (i.e. filling in the second param for adDisplayContainer)
           //also doesn't not seem to work nicely with podded ads if you don't use it.
-          _IMAAdDisplayContainer = new google.ima.AdDisplayContainer(_amc.ui.adWrapper[0],
+
+          //for IMA, we always want to use the plugins element to house the IMA UI. This allows it to behave
+          //properly with the Alice skin.
+          _IMAAdDisplayContainer = new google.ima.AdDisplayContainer(_amc.ui.pluginsElement[0],
```
angieisrani: Should all ad managers do this or do others use adWrapper?  Should pluginsElement be the new adWrapper?

aeng7: IMA is special in that it controls the ad UI and provides the capability for us to provide to them a container to store this UI and a container to place the video. I believe we should only use the plugins element (instead of the ad wrapper) for situations like these. Our other ad managers behave differently from IMA and should still use ad wrapper to house the video. We control the UI for our other ad managers, at least the current ones.

angieisrani:
LGTM

PR #37, opened 2016-01-11
[PBI-731] Implement a timeout for fw_onAdRequestComplete event

dchauOoyala:
- Adds a timeout to check if there is a response for the ad list request
- Raises error if there is no response

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -278,8 +278,8 @@ OO.Ads.manager(function(_, $) {
       _clearAdRequestTimeout();
       if (!fwContext._adResponse) {
         var error = "ad request timeout";
+        fw_onError(error);
         OO.log("FW: freewheel ad request timeout");
-        amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
         slotEndedCallbacks[adRequestType]();
         delete slotEndedCallbacks[adRequestType];
       }
@@ -296,7 +296,7 @@ OO.Ads.manager(function(_, $) {
     var _setAdRequestTimeout = _.bind(function(callback, duration){
       if (adRequestTimeout) {
         var error = "Ad Request Timeout already exists - bad state";
-        amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
+        fw_onError(error);
       } else {
         adRequestTimeout = _.delay(callback, duration);
       }
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -269,8 +269,11 @@ OO.Ads.manager(function(_, $) {
           OO.log("FW: freewheel ad request timeout");
           amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
         }
+        console.log("THE TIME: ");
+        console.log(ad_request_timer.end_time - ad_request_timer.start_time);
       });
 
+      ad_request_timer.start_time = (new Date()).getTime();
       fwContext.submitRequest();
       fwAdDataRequested = true;
       setAdRequestTimeout(adRequestTimeout, amc.MAX_AD_REQUEST_TIMEOUT);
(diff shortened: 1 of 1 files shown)


pilievOoyala:
The timeout needs to be clear on successful callback and when we delete/destroy the plugin.

js/freewheel.js
```diff
@@ -261,9 +261,31 @@ OO.Ads.manager(function(_, $) {
...
+          OO.log("FW: freewheel ad request timeout");
+          amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
+          slotEndedCallbacks[adRequestType]();
+          delete slotEndedCallbacks[adRequestType];
+        }
+      }, this);
+
       fwContext.submitRequest();
       fwAdDataRequested = true;
+      setAdRequestTimeout(adRequestTimeout, amc.MAX_AD_REQUEST_TIMEOUT);
```
pilievOoyala: If the adRequestTimeout already exists, then we are in a bad state.  So throw error.

js/freewheel.js
```diff
@@ -261,9 +263,39 @@ OO.Ads.manager(function(_, $) {
...
+     * if we timeout.
+     * @private
+     * @method Freewheel#_setAdRequestTimeout
+     * @param callback The function to call when we time out
+     * @param duration the time to wait before timing out
+     */
+    var _setAdRequestTimeout = _.bind(function(callback, duration){
+      if (adRequestTimeout) {
+        var error = "Ad Request Timeout already exists - bad state";
+        amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
```
pilievOoyala: You've used this in a couple of places. If you want, you can put this amc.raiseAdError into a function. That way you guarantee the format of the error message will always be the same.  Just a suggestion, not required.

js/freewheel.js
```diff
@@ -283,6 +315,8 @@ OO.Ads.manager(function(_, $) {
       } else {
         OO.log("FW: freewheel metadata request failure");
       }
+      // clear ad request timeout since fw_onAdRequestComplete was called
+      _clearAdRequestTimeout();
```
pilievOoyala: Can you move this to the top of the function?  That way if someone ads code and it happens to exit early, we don't forget to do this.

js/freewheel.js
```diff
@@ -261,9 +263,39 @@ OO.Ads.manager(function(_, $) {
...
+      // Set up function to log error if ad response is null after maximum duration allowed for ad request to respond
+      // has exceeded.
+      var adRequestTimeout = _.bind(function(){
+        if (!fwContext._adResponse) {
+          var error = "ad request timeout";
+          OO.log("FW: freewheel ad request timeout");
+          amc.raiseAdError("FW: An ad error has occurred. The error string reported was: " + error);
+          slotEndedCallbacks[adRequestType]();
+          delete slotEndedCallbacks[adRequestType];
+          _clearAdRequestTimeout();
```
pilievOoyala: Can you move the clear to be outside the if.  We always want to clear it.

js/freewheel.js
```diff
@@ -261,9 +263,39 @@ OO.Ads.manager(function(_, $) {
         ].join('')
       ];
 
-      // TODO: Implement a timeout here in case fw_onAdRequestComplete doesn't get called - PBI-731
+      // Set up function to log error if ad response is null after maximum duration allowed for ad request to respond
+      // has exceeded.
+      var adRequestTimeout = _.bind(function(){
```
pilievOoyala: Is it possible to move this function to be defined out side. In general we are trying to not define functions within functions. (Easier to debug.)

pilievOoyala:
Good work, just a couple of style notes to avoid potential problems in the future.

pilievOoyala:
Looks good to me.  Just make sure to validate that the timeout and all the code won't crash the player and we can still watch the video.

pilievOoyala:
We talked and you validated on desktop and mobile.  Good work!

dchauOoyala:
Hooray!

PR #39, opened 2016-01-14
Pbw 4052

aeng7:
-IMA will now auto scale its ads to fit its container. This is to support the change in PBW-4052 where the ad video space is smaller when the ad control bar is "always on."

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -697,7 +697,7 @@ require("../html5-common/js/utils/utils.js");
 
             var w = _amc.ui.width;
             var h = _amc.ui.height;
-            _IMAAdsManager.init(w, h, google.ima.ViewMode.NORMAL);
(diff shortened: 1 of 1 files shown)


js/google_ima.js
```diff
@@ -776,19 +774,7 @@ require("../html5-common/js/utils/utils.js");
...
-            // of the slot. We also make the slot a bit shorter, so there is
-            // a padding at the bottom.
-            _IMAAdsManager.resize(_amc.ui.width, _amc.ui.height - OVERLAY_OFFSET, viewMode);
-          }
-          else
-          {
-            // For linear ad, set the size to the full video player size.
-            _IMAAdsManager.resize(_amc.ui.width, _amc.ui.height, viewMode);
-          }
+          _IMAAdsManager.resize(google.ima.AdsRenderingSettings.AUTO_SCALE, google.ima.AdsRenderingSettings.AUTO_SCALE, viewMode);
```
aeng7: I tested with overlays and did not see any issues when not using OVERLAY_OFFSET. This is most likely due to the fact that the control bar will not overlap with IMA ads anymore.

aeng7:
Closing PR. Some additional changes required so that "always on" mode affects only IMA.

PR #40, opened 2016-01-15
Pbw 4052

aeng7:
-IMA will now automatically scale their ads to fit the ad display container
-IMA will use the new playerSkinPluginsElement for the ad display container


aeng7:
Please merge the same time as mjolnir PR 473

PR #41, opened 2016-01-19
[PBW-4052]

aeng7:
-When in single video element mode, IMA will now wait until content resume message from IMA before ending ad pod.


pilievOoyala:
lgtm

PR #42, opened 2016-01-20
[PBI-1363] Updated code to support v4 changes, Overlay will need to b…

gfrankOoyala:
…e fixed later.

### DIFF
--- a/js/ad_manager_videoplaza.js
+++ b/js/ad_manager_videoplaza.js
@@ -425,10 +425,10 @@ OO.Ads.manager(function(_, $) {
 
       if (ad.isLinear) {
         this.lastVideoAd = ad;
+        this.tracker.trackEvent(ad.ad.creatives[0], videoplaza.tracking.Tracker.CreativeEventType.START);
         this.amc.notifyPodStarted(ad.id, 1);
         adCompletedCallback = _.bind(function(amc, adId) {
-          amc.notifyLinearAdEnded(adId);
-          amc.notifyPodEnded(adId);
+          notifyAmcOfAdEnd(adId);
         }, this, this.amc, ad.id);
         this.checkCompanionAds(ad.ad);
         this.amc.showSkipVideoAdButton(true);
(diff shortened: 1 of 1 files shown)


js/ad_manager_videoplaza.js
```diff
@@ -464,56 +420,36 @@ OO.Ads.manager(function(_, $) {
...
-        this.quartileTracks = {};
-        ui.adVideoElement[0].play();
         this.lastVideoAd = ad;
-        this.tracker.trackEvent(ad.ad.creatives[0], videoplaza.tracking.Tracker.CreativeEventType.START);
-        adPodStartedCallback();
-        ui.rootElement.find('div.oo_tap_panel').css('display', 'block');
+        this.amc.notifyPodStarted(ad.id, 1);
+        adCompletedCallback = _.bind(function(amc, adId) {
+          amc.notifyLinearAdEnded(adId);
+          amc.notifyPodEnded(adId);
```
aeng7: Would it prove beneficial to have a function to encompass the functionality of notifying of a linear ad pod end? There is another instance of these two functions in cancelAd.

js/ad_manager_videoplaza.js
```diff
@@ -527,13 +463,17 @@ OO.Ads.manager(function(_, $) {
...
     this.cancelAd = function(ad) {
-      //TODO: cancel linear ads
-      var overlayContainer = this.amc.ui.rootElement.find(this.OVERLAY_AD_CONTAINER);
-      if (overlayContainer) {
-        overlayContainer.remove();
-        this.lastOverlayAd = null;
-        if (_.isFunction(this.adEndedAMCCallback)) {
-          this.adEndedAMCCallback();
+      if (!this.amc || !this.amc.ui) {
+        return;
```
aeng7: How come we want to return if the amc ui does not exist?

js/ad_manager_videoplaza.js
```diff
@@ -464,56 +420,36 @@ OO.Ads.manager(function(_, $) {
...
-          "text-align": "center"
-        });
-        this.checkCompanionAds(ad.ad);
-        this.quartileTracks = {};
-        ui.adVideoElement[0].play();
         this.lastVideoAd = ad;
-        this.tracker.trackEvent(ad.ad.creatives[0], videoplaza.tracking.Tracker.CreativeEventType.START);
-        adPodStartedCallback();
-        ui.rootElement.find('div.oo_tap_panel').css('display', 'block');
+        this.amc.notifyPodStarted(ad.id, 1);
```
aeng7: Should we return or have some error handling if the ad argument is null/undefined?

js/ad_manager_videoplaza.js
gfrankOoyala: Because amc ui contains a lot of variables that are needed like startTime, so if it is null that means AMC is broken or initialized properly.

aeng7: Got it, it just looked somewhat out of place to me because cancelAd didn't make use of it.

aeng7:
LGTM!

PR #45, opened 2016-01-28
Pbi 742

aeng7:
Implemented Vast 3.0 skipoffset functionality.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -75,15 +75,15 @@ OO.Ads.manager(function(_, $) {
                         nonLinear: { tracking: {} } };
     var adCompletedCallback = null;
 
-    var VERSION_2_0 = '2.0';
-    var VERSION_3_0 = '3.0';
-    var SUPPORTED_VERSIONS = [VERSION_2_0, VERSION_3_0];
+    var VERSION_MAJOR_2 = '2';
+    var VERSION_MAJOR_3 = '3';
+    var SUPPORTED_VERSIONS = [VERSION_MAJOR_2, VERSION_MAJOR_3];
     var FEATURES = {
       SKIP_AD : "skipAd"
     };
     var SUPPORTED_FEATURES = {};
-    SUPPORTED_FEATURES[VERSION_2_0] = [];
-    SUPPORTED_FEATURES[VERSION_3_0] = [FEATURES.SKIP_AD];
+    SUPPORTED_FEATURES[VERSION_MAJOR_2] = [];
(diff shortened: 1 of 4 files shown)


aeng7:
Please do not merge until Mjolnir PR 486 has been merged.

aeng7:
Update to pass seconds rather than ms to AMC via showSkipVideoAdButton

js/ad_manager_vast.js
```diff
@@ -93,12 +104,47 @@ OO.Ads.manager(function(_, $) {
...
+    }, this);
+
+    /**
+     * Checks to see if this ad manager supports a given Vast version.
+     * @private
+     * @method Vast#supportsVersion
+     * @returns {boolean} true if the version is supported by this ad manager, false otherwise
+     */
+    var supportsVersion = _.bind(function(version) {
+      return _.contains(SUPPORTED_VERSIONS, version);
```
pilievOoyala: I believe there is a version 2.1. I remember encountering this in the past and we had parsing errors.

aeng7: Got it, I'll add logic for supported Vast versions to check major versions (and minor if necessary).

js/ad_manager_vast.js
```diff
@@ -305,6 +349,46 @@ OO.Ads.manager(function(_, $) {
     };
 
     /**
+     * Determine if a Vast ad is skippable, and if so, when the skip ad button should be displayed.
+     * Notifies AMC of the result.
+     * @private
+     * @method Vast#calculateSkipAdOffset
+     * @param {object} adWrapper The current Ad's metadata.
+     */
+    var calculateSkipAdOffset = _.bind(function(adWrapper) {
```
pilievOoyala: when it's name starts with calculate, it makes me think it should return a value. Maybe name it initSkipAdOffset?

aeng7: Good point! Will do

js/ad_manager_vast.js
```diff
@@ -740,6 +825,7 @@ OO.Ads.manager(function(_, $) {
      * @private
      * @method Vast#parseLinearAd
      * @param {xml} Xml containing the ad data to be parsed.
+     * @param {number} version The Vast version
```
pilievOoyala: Is this optional? I doesn't look like it being passed into the function ebcause it's only got one param

aeng7: I think this is a remnant of a previous checkin. Will remove. Nice catch!

aeng7:
Included Peter's suggestions.

pilievOoyala:
Looks good.

PR #47, opened 2016-02-08
Pbi 741a

aeng7:
-Vast 3.0 podded ads functionality
-supports linear podded ads (tested with a sample tag) and theoretically supports podded ads that end with a nonlinear ad (untested due to lack of test tag)
-removed some wrapper ads functionality and unit test due to them functioning incorrectly before. Will have to be re-added once we tackle wrapper ads

### DIFF
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -130,6 +130,7 @@ OO.Ads.manager(function(_, $) {
      * @public
      * @param {object} ad The ad object to cancel
      * @param {object} params An object containing information about the cancellation
+     *                 code : The amc.AD_CANCEL_CODE for the cancellation
      */
     this.cancelAd = function(ad, params) {
     };
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -129,7 +129,8 @@ OO.Ads.manager(function(_, $) {
      * @method AdManager#cancelAd
      * @public
      * @param {object} ad The ad object to cancel
-     * @param {object} params An object containing information about the cancellation
+     * @param {object} params An object containing information about the cancellation. It will include the
+     *                        following fields:
      *                 code : The amc.AD_CANCEL_CODE for the cancellation
      */
     this.cancelAd = function(ad, params) {
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -216,8 +216,10 @@ OO.Ads.manager(function(_, $) {
      * this error itself, it can use this time to end the ad playback.
      * @method AdManager#adVideoError
      * @public
+     * @param {object} adWrapper The current Ad's metadata
+     * @param {number} errorCode The error code associated with the video playback error
      */
-    this.adVideoError = function() {
+    this.adVideoError = function(adWrapper, errorCode) {
     };
 
     /**
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -198,6 +198,15 @@ OO.Ads.manager(function(_, $) {
     this.playerClicked = function(amcAd, showPage) {
     };
 
+    /**
+     * <i>Optional.</i><br/>
+     * Called when the player detects start of ad video playback.
+     * @method AdManager#adVideoPlaying
+     * @public
+     */
+    this.adVideoPlaying = function() {
+
+    };
 
     /**
      * <i>Optional.</i><br/>
@@ -222,7 +231,7 @@ OO.Ads.manager(function(_, $) {
     var _onContentChanged = function() {
       // Callback for example listener registered in this.initialize
     }
-  }
+  };
 
   return new AdManager();
 });
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -129,8 +129,9 @@ OO.Ads.manager(function(_, $) {
      * @method AdManager#cancelAd
      * @public
      * @param {object} ad The ad object to cancel
+     * @param {object} params An object containing information about the cancellation
      */
-    this.cancelAd = function(ad) {
+    this.cancelAd = function(ad, params) {
     };
 
     /**
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -449,16 +449,21 @@ OO.Ads.manager(function(_, $) {
      * @public
      * @method Vast#cancelAd
      * @param {object} ad The Ad that needs to be cancelled
+     * @param {object} params An object containing information about the cancellation
      */
-    this.cancelAd = function(ad) {
+    this.cancelAd = function(ad, params) {
       //TODO: add timout logic if needed here as well.
       if (!this.amc || !this.amc.ui || !ad) {
         return;
       }
-      if (ad.isLinear) {
-        this.adVideoEnded();
+      if(params && params.code === this.amc.AD_CANCEL_CODE.TIMEOUT) {
+        failedAd();
       } else {
-        this.endAd(ad);
+        if (ad.isLinear) {
+          this.adVideoEnded();
+        } else {
+          this.endAd(ad);
+        }
       }
     };
 
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -1112,17 +1112,17 @@ OO.Ads.manager(function(_, $) {
       } else {
         var ad;
         //TODO: Determine when we show standalone ads if podded ads are available
-        //Show podded ads if available
-        if(!_.isEmpty(vastAds.podded)) {
-          handleAds(vastAds.podded, adLoaded);
-        }
-        //else show first standalone ad
-        else {
+        //If there are no podded ads
+        if(_.isEmpty(vastAds.podded)) {
+          //show the first standalone ad
           ad = vastAds.standalone[0];
+          if (ad) {
+            handleAds([ad], adLoaded);
+          }
         }
-
-        if (ad) {
-          handleAds([ad], adLoaded);
+        //else show the podded ads
+        else {
+          handleAds(vastAds.podded, adLoaded);
         }
       }
     };
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -59,17 +59,19 @@ OO.Ads.manager(function(_, $) {
     this.VAST_AD_CONTAINER = '#vast_ad_container';
     this.currentAdBeingLoaded = null;
     var adCompletedCallback = null;
+    var currentAd = null;
 
     var VERSION_MAJOR_2 = '2';
     var VERSION_MAJOR_3 = '3';
     var SUPPORTED_VERSIONS = [VERSION_MAJOR_2, VERSION_MAJOR_3];
     var FEATURES = {
       SKIP_AD : "skipAd",
-      PODDED_ADS : "poddedAds"
+      PODDED_ADS : "poddedAds",
+      AD_FALLBACK : "adFallback"
     };
     var SUPPORTED_FEATURES = {};
     SUPPORTED_FEATURES[VERSION_MAJOR_2] = [];
-    SUPPORTED_FEATURES[VERSION_MAJOR_3] = [FEATURES.SKIP_AD, FEATURES.PODDED_ADS];
+    SUPPORTED_FEATURES[VERSION_MAJOR_3] = [FEATURES.SKIP_AD, FEATURES.PODDED_ADS, FEATURES.AD_FALLBACK];
 
     var AD_TYPE = {
       INLINE : "InLine",
@@ -170,6 +172,18 @@ OO.Ads.manager(function(_, $) {
     }, this);
 
     /**
+     * Checks to see if the given Vast version supports the ad fallback functionality, as per Vast specs
+     * for different versions.
+     * @private
+     * @method Vast#supportsAdFallback
+     * @returns {boolean} true if the ad fallback functionality is supported in the specified Vast version,
+     *                    false otherwise
+     */
+    var supportsAdFallback = _.bind(function(version) {
+      return _.contains(SUPPORTED_FEATURES[getMajorVersion(version)], FEATURES.AD_FALLBACK);
+    }, this);
+
+    /**
      * Default template to use when creating the vast ad object.
      * @private
      * @method Vast#getVastTemplate
@@ -248,8 +262,10 @@ OO.Ads.manager(function(_, $) {
      * @method vast#failedAd
      */
     var failedAd = _.bind(function() {
-      // TODO: Do not destroy the whole ad manager if one ad fails!
-      this.destroy();
+      if (currentAd && currentAd.ad.fallbackAd) {
+        addToTimeline(currentAd.ad.fallbackAd);
+      }
+      this.cancelAd(currentAd);
     }, this);
 
     /**
@@ -307,6 +323,18 @@ OO.Ads.manager(function(_, $) {
     };
 
     /**
+     * Called when the ad starts playback.
+     * @public
+     * @method Vast#adVideoPlaying
+     */
+    this.adVideoPlaying = function() {
+      //Push the next ad in the ad pod to the timeline
+      if (currentAd && currentAd.ad.nextAdInPod) {
+        addToTimeline(currentAd.ad.nextAdInPod);
+      }
+    };
+
+    /**
      * When the ad is finished playing we need to call the AMC callback that was provided to let the AMC know that the
      * ad is finished playing.
      * @public
@@ -326,7 +354,6 @@ OO.Ads.manager(function(_, $) {
      * @method Vast#adVideoError
      */
     this.adVideoError = function() {
-      this.adVideoEnded();
       // VTC will pause the ad when the video element loses focus
       failedAd();
     };
@@ -343,6 +370,7 @@ OO.Ads.manager(function(_, $) {
      * @param {object} adWrapper The current Ad's metadata
      */
     this.playAd = function(adWrapper) {
+      currentAd = adWrapper;
       // When the ad is done, trigger callback
       if (adWrapper.isLinear) {
         if (adWrapper.ad.adPodIndex === 1) {
@@ -350,7 +378,7 @@ OO.Ads.manager(function(_, $) {
         }
         adCompletedCallback = _.bind(function(ad) {
             this.endAd(ad);
-          adCompletedCallback = null;
+            adCompletedCallback = null;
           }, this, adWrapper);
         this.checkCompanionAds(adWrapper.ad);
         initSkipAdOffset(adWrapper);
@@ -454,6 +482,7 @@ OO.Ads.manager(function(_, $) {
           this.amc.notifyNonlinearAdEnded(ad.id);
         }
       }
+      currentAd = null;
     };
 
     /**
@@ -476,10 +505,10 @@ OO.Ads.manager(function(_, $) {
      * @private
      * @method Vast#addToTimeline
      * @param {object} metadata The ad metadata that is being added to the timeline.
-     * @param {object} adLoaded The ad object that has been loaded.
+     * @param {boolean} True if the ad was added to the timeline successfully, false otherwise
      */
-    var addToTimeline = _.bind(function(metadata, adLoaded) {
-      if (!metadata) return;
+    var addToTimeline = _.bind(function(metadata) {
+      if (!metadata) return false;
       var timeline = [];
       var type, duration;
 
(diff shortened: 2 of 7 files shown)


aeng7:
Please merge after Mjolnir PR 497 has been merged

aeng7:
Tested mocked XML with non-linear ad at the end of podded ad. Worked fine.

aeng7:
Do we want to add logic where if we find a non-linear overlay in the middle of a podded ad sequence, we push it to the end? The Vast 3.0 spec explicitly says that the non-linear overlay must be at the end of a podded ad sequence.

aeng7:
Will add modified logic to handle ending ad pod before a non linear overlay if one exists. Also will change it so that ad count in pod will not count non linear ads. Will also add unit tests for these. Please do not merge yet.

aeng7:
Ok to merge after review.

js/ad_manager_vast.js
```diff
@@ -935,41 +993,113 @@ OO.Ads.manager(function(_, $) {
...
+     * @method Vast#handleAds
+     * @param {Array} ads An array of ad objects
+     * @param {object} adLoaded The ad loaded object and metadata
+     */
+    var handleAds = _.bind(function(ads, adLoaded) {
+      //find out how many non linear ads we have so as to not count them
+      //when determining ad pod length
+      var nonLinearAdCount = 0;
+      _.each(ads, _.bind(function(ad) {
+        if (!_.isEmpty(ad.nonLinear)) {
```
dchauOoyala: I think this will always evaluate to "not empty" because ad.nonLinear will always have "tracking" as a property like so:
```
ad.nonLinear = {
  tracking: {}
}
```
I think this happens because of LINE 789. double check me on that.

js/ad_manager_vast.js
```diff
@@ -722,29 +769,28 @@ OO.Ads.manager(function(_, $) {
...
+        ad.linear.ClickTracking = wrapperAds.linear.ClickTracking.concat(ad.linear.ClickTracking || []);
+      }
+      if (wrapperAds.linear.tracking) {
+        if (!ad.linear.tracking) { ad.linear.tracking  = {}; }
+        _.each(wrapperAds.linear.tracking, function(value, key) {
+          ad.linear.tracking[key] = ad.linear.tracking[key] ? value.concat(ad.linear.tracking[key]) : value;
+        });
+      }
+      if (wrapperAds.nonLinear.tracking) {
+        if (!ad.nonLinear.tracking) { ad.nonLinear.tracking = {}; }
```
dchauOoyala: This is in reference to comment on LINE 1071.

js/ad_manager_vast.js
```diff
@@ -935,41 +993,113 @@ OO.Ads.manager(function(_, $) {
...
+        podded : [],
+        standalone : []
+      };
+      //parse the ad objects from the XML
+      var ads = this.parseAds(vastXML);
+      //check to see if any ads are sequenced (are podded)
+      _.each(ads, _.bind(function(ad) {
+        if (supportsPoddedAds(ad.version) && typeof ad.sequence !== 'undefined' && _.isNumber(parseInt(ad.sequence))) {
+          //Assume sequences will start from 1
+          result.podded[+ad.sequence - 1] = ad;
```
dchauOoyala: What does the "+" prefix do in the context of the array access?

js/ad_manager_vast.js
```diff
@@ -935,41 +993,113 @@ OO.Ads.manager(function(_, $) {
...
+            linear: { tracking: {}, ClickTracking: [] },
+            nonLinear: { tracking: {} } };
+          var params = {
+            adPodIndex : index + 1,
+            adPodLength : ads.length - nonLinearAdCount
+          };
+          this._mergeVastAdResult(ad, wrapperAds);
+          if (this._handleLinearAd(ad, adLoaded, params) || this._handleNonLinearAd(ad, adLoaded, params)) {
+            this.loaded = true;
+            this.trigger(this.READY, this);
```
dchauOoyala: Not sure, but if this is an ad pod should `` this.trigger(this.READY, this); `` be called after all the ads have been added to the time line and not after a single iteration?

js/ad_manager_vast.js
```diff
@@ -697,24 +739,29 @@ OO.Ads.manager(function(_, $) {
...
       // filter our playable stream:
-      var firstNonLinearAd = _.find(this.inlineAd.ads, function(v){ return !_.isEmpty(v.nonLinear.url); }, this);
-      if (!firstNonLinearAd) { return false; }
-      var adURL = firstNonLinearAd.nonLinear.url;
-      this.vastAdUnit.streamUrl = adURL;
-      _.extend(this.vastAdUnit.data, firstNonLinearAd);
-      this.vastAdUnit.data.tracking = firstNonLinearAd.nonLinear.tracking;
-
-      if (this.vastAdUnit.streamUrl == null) {
+      if (_.isEmpty(ad.nonLinear.url)) { return false; }
```
dchauOoyala: Nitpicky, but for readability/debugging purposes I think t he return should be on a newline

js/ad_manager_vast.js
```diff
@@ -697,24 +739,29 @@ OO.Ads.manager(function(_, $) {
...
+      params = params ? params : {};
+      var adURL = ad.nonLinear.url;
+      var vastAdUnit = { data: {}, vastUrl: this.vastUrl, maxBitrateStream: null };
```
dchauOoyala: Nitpicky, but I think it's slightly more readable to declare the object literal like:
```
var vastAdUnit = {
  data: {},
  vastUrl: this.vastUrl,
  maxBitrateStream: null
};
```

js/ad_manager_vast.js
aeng7: The mergeVastAdResult function gets called inside handleAds a few lines down. ad.nonLinear is empty at this time.

js/ad_manager_vast.js
aeng7: Agreed, will make this change.

js/ad_manager_vast.js
aeng7: It turns it into a number. ad.sequence is originally a string at this point.

js/ad_manager_vast.js
aeng7: Yep, sounds like it should be, but I don't believe it does anything here. Greg did want us to try to remove the dependency on the Emitter object since it doesn't work fully. I'll see if I could do that in this branch. If not will follow up on this.

js/ad_manager_vast.js
```diff
@@ -979,63 +1109,25 @@ OO.Ads.manager(function(_, $) {
...
-        } else {
-          OO.log("Max wrapper depth reached.", this.currentDepth, OO.playerParams.maxVastWrapperDepth);
-          this.errorType = "tooManyWrapper";
-          this.trigger(this.ERROR, this);
-          failedAd();
+      } else {
+        var ad;
+        //TODO: Determine when we show standalone ads if podded ads are available
+        //Show podded ads if available
+        if(!_.isEmpty(vastAds.podded)) {
```
dchauOoyala: I feel this conditional along with the conditional on LINE 1129 might read off better like:
```
if(_.isEmpty(vastAds.podded)) {
	ad = vastAds.standalone[0];
	if (ad) {
		handleAds([ad], adLoaded);
	}
}
else {
	handleAds(vastAds.podded, adLoaded);
}
```
that way the "branch"/flow for the standalone logic is grouped together.

js/ad_manager_vast.js
dchauOoyala: Ahh okay, oops

js/ad_manager_vast.js
dchauOoyala: Ahh, I see

js/ad_manager_vast.js
aeng7: Will make this change.

js/ad_manager_template.js
```diff
@@ -198,6 +199,15 @@ OO.Ads.manager(function(_, $) {
...
     };
 
+    /**
+     * <i>Optional.</i><br/>
+     * Called when the player detects start of ad video playback.
+     * @method AdManager#adVideoPlaying
+     * @public
+     */
+    this.adVideoPlaying = function() {
+
```
gfrankOoyala: Why do we need this if it is empty?

js/ad_manager_vast.js
```diff
@@ -94,11 +87,20 @@ OO.Ads.manager(function(_, $) {
       'close' ];
 
     /**
-     * Helper function to verify that the xml has a valid vast ad in it and that it is a valid xml.
+     * Helper function to verify that XML is valid
      * @param {xml} vastXML Contains the vast ad data to be parsed.
      * @returns {boolean} Returns true if the xml is valid otherwise it returns false.
      */
-    var isValidVastXML = _.bind(function(vastXML) {
+    this.isValidVastXML = _.bind(function(vastXML) {
```
gfrankOoyala: Public functions don't need _.bind

js/ad_manager_template.js
aeng7: This is the ad manager template and will be used as a reference for people who want to implement their own ad managers.

js/ad_manager_vast.js
```diff
@@ -406,22 +449,45 @@ OO.Ads.manager(function(_, $) {
...
+    };
+
+    /**
+     *
+     * @param ad
+     */
+    this.endAd = function(ad) {
       if (ad) {
         if (ad.isLinear) {
+          var endOfPod = ad.ad.adPodIndex === ad.ad.adPodLength;
```
gfrankOoyala: Can me make this an if instead, makes it easier to read.

js/ad_manager_vast.js
aeng7: Got it, will make this change.

js/ad_manager_vast.js
```diff
@@ -442,48 +509,33 @@ OO.Ads.manager(function(_, $) {
...
       {
         type = this.amc.ADTYPE.NONLINEAR_OVERLAY;
-        duration = ad.data.nonLinear.Duration ?  OO.timeStringToSeconds(ad.data.nonLinear.Duration) : 0;
+        duration = metadata.data.nonLinear.Duration ?  OO.timeStringToSeconds(metadata.data.nonLinear.Duration) : 0;
       }
-      var positionSeconds = adLoaded.time/1000;
 
-      // Save the stream data for use by VideoController
-      var streams = {};
-      var linearAd = ad.data.linear;
```
gfrankOoyala: Why was this removed? We need to build the streamData object ot have all the supported streams listed so the VTC can pick the right one.

js/ad_manager_vast.js
```diff
@@ -636,11 +688,10 @@ OO.Ads.manager(function(_, $) {
     this._onVastError = function() {
       this.errorType = 'directAjaxFailed';
       this._ajax(this._getProxyUrl(), this._onFinalError, 'script');
-      this.trigger(this.ERROR, this);
```
gfrankOoyala: Can we log an error here please

js/ad_manager_vast.js
```diff
@@ -648,7 +699,6 @@ OO.Ads.manager(function(_, $) {
      */
     this._onFinalError = function() {
       this.errorType = "proxyAjaxFailed";
-      this.trigger(this.ERROR, this);
```
gfrankOoyala: log here as well

js/ad_manager_vast.js
```diff
@@ -935,41 +1018,171 @@ OO.Ads.manager(function(_, $) {
...
+        }, this));
+
+        handled = addToTimeline(adUnits[0]);
+      }
+
+      if (handled) {
+        this.loaded = true;
+      } else {
+        this.errorType = "noAd";
+        failedAd();
```
gfrankOoyala: Can we log when this happens.

js/ad_manager_vast.js
aeng7: This was moved to inside the _handleLinearAd function.

js/ad_manager_vast.js
aeng7: This will be part of the error handling code that Douglas will be checking in. Should I include a temporary one for now?

js/ad_manager_vast.js
aeng7: Will do.

aeng7:
I've found a bug with replays, please do not merge yet.

js/ad_manager_template.js
```diff
@@ -129,8 +129,9 @@ OO.Ads.manager(function(_, $) {
      * @method AdManager#cancelAd
      * @public
      * @param {object} ad The ad object to cancel
+     * @param {object} params An object containing information about the cancellation
```
angieisrani: What kind of information?  Is that documented anywhere?

js/ad_manager_vast.js
```diff
@@ -12,7 +12,6 @@ require("../html5-common/js/utils/InitModules/InitOOPlayerParamsDefault.js");
 
 require("../html5-common/js/utils/constants.js");
 require("../html5-common/js/utils/utils.js");
-require("../html5-common/js/classes/emitter.js");
```
angieisrani: woo!

js/ad_manager_vast.js
```diff
@@ -40,23 +37,18 @@ OO.Ads.manager(function(_, $) {
...
    * @property {object} lastOverlayAd Contains the ad information for the overlay that was displayed before it was removed.
    * This is used so we know what to add back to the screen after the video ad is done and the main video hasn't ended.
-   * @property {string} ERROR Constant used when triggering an error to indicate it was a Vast error
-   * @property {string} READY Constant used to trigger an event to indicate that the vast ad is ready for use
    * @property {string} VAST_AD_CONTAINER Constant used to keep track of the Vast Ad container div/layer that is used to
    * show the ads
    * @property {object} currentAdBeingLoaded Stores the ad data of the ad that is currently being loaded
-   * @property {object} wrapperAds Is used to keep track of the analytic and clickthrough info of an ad
    */
-  var Vast = OO.inherit(OO.Emitter, function() {
```
angieisrani: woo!

js/ad_manager_vast.js
```diff
@@ -235,8 +264,32 @@ OO.Ads.manager(function(_, $) {
...
      */
     var failedAd = _.bind(function() {
-      // TODO: Do not destory the whole ad manager if one ad fails!
-      this.destroy();
+      var metadata = null;
+      var badAd = currentAd;
+      currentAd = null;
+
+      if (badAd) {
+        if(badAd.ad.fallbackAd) {
```
angieisrani: might need to check if badAd.ad?

js/ad_manager_vast.js
```diff
@@ -294,6 +347,22 @@ OO.Ads.manager(function(_, $) {
     };
 
     /**
+     * Called when the ad starts playback.
+     * @public
+     * @method Vast#adVideoPlaying
+     */
+    this.adVideoPlaying = function() {
+      if (currentAd && currentAd.ad.nextAdInPod) {
```
angieisrani: if currentAd.ad?

js/ad_manager_vast.js
```diff
@@ -294,6 +347,22 @@ OO.Ads.manager(function(_, $) {
...
+        var metadata = currentAd.ad.nextAdInPod;
+        var ad = generateAd(metadata);
+
+        if (metadata.streamUrl != null || (ad.adType == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(metadata.streams))) {
```
angieisrani: if metadata?

angieisrani: What is the difference between streamUrl and streams

js/ad_manager_vast.js
```diff
@@ -235,8 +264,32 @@ OO.Ads.manager(function(_, $) {
...
+      if (badAd) {
+        if(badAd.ad.fallbackAd) {
+          metadata = badAd.ad.fallbackAd;
+        }
+        //notify amc of the end of the failed ad
+        if (badAd.isLinear) {
+          this.amc.notifyLinearAdEnded(badAd.id);
+        } else {
+          this.lastOverlayAd = null;
+          this.amc.notifyNonlinearAdEnded(badAd.id);
```
angieisrani: Can you call adCompletedCallback?

js/ad_manager_vast.js
```diff
@@ -406,22 +486,43 @@ OO.Ads.manager(function(_, $) {
...
+        if (ad.isLinear) {
+          this.adVideoEnded();
+        } else {
+          this.endAd(ad);
+        }
+      }
+    };
+
+    /**
+     *
```
angieisrani: needs jsdoc
Is this public because AMF is calling it?  Does it need to be public?

angieisrani: same for the functions that were made public

js/ad_manager_vast.js
```diff
@@ -671,80 +770,101 @@ OO.Ads.manager(function(_, $) {
      *  If a linear ad is found, then it is parsed and sent to be added to the time via addToTimeLine.
      * @public
      * @method Vast#_handleLinearAd
+     * @param {object} ad The ad object
      * @param {object} adLoaded The ad that was loaded
-     * @returns {boolean} True if the ad was loaded and a stream was found; else false.
+     * @param {object} params Additional params
```
angieisrani: what kind of params

js/ad_manager_vast.js
```diff
@@ -671,80 +770,101 @@ OO.Ads.manager(function(_, $) {
...
     };
 
     /**
      * If a non-linear Ad is found then it is parsed and added to the timeline via the addToTimeline function.
      * @public
      * @method Vast#_handleNonLinearAd
+     * @param {object} ad The ad object
      * @param {object} adLoaded The ad that was loaded
-     * @returns {boolean} True if the load was successful and a stream was found otherwise false.
+     * @param {object} params Additional params
```
angieisrani: what kinds

js/ad_manager_vast.js
```diff
@@ -671,80 +770,101 @@ OO.Ads.manager(function(_, $) {
...
+     * @returns {object} The ad unit object ready to be added to the timeline
      */
-    this._handleNonLinearAd = function(adLoaded) {
+    this._handleNonLinearAd = function(ad, adLoaded, params) {
```
angieisrani: this function is public but is preceded by an underscore.  Is it meant to have been private?

js/freewheel.js
```diff
@@ -460,9 +460,9 @@ OO.Ads.manager(function(_, $) {
             slotEndedCallbacks[ad.ad.getCustomId()] = _.bind(function(adId) {
                 amc.notifyPodEnded(adId);
               }, this, ad.id);
-            adStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(details) {
-                amc.notifyLinearAdStarted(this.name, details);
-              }, this);
+            adStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(adId, details) {
```
angieisrani: is this new parameter already being used?  i don't see any other changes such as to the function calls

test/unit-test-helpers/mock_amc.js
```diff
@@ -20,6 +20,12 @@ fake_amc = function() {
     LINEAR_VIDEO : "linearVideo",
     COMPANION : "companion"
   };
+  this.AD_CANCEL_CODE = {
+    SKIPPED : "skipped",
+    TIMEOUT : "timeout",
+    ERROR : "error",
+    STREAM_ENDED : "streamEnded"
+  };
```
angieisrani: i am assuming this was already defined in real AMC?

aeng7: Yep, it's defined in the pbi-741 branch, which has not been merged yet.

js/ad_manager_vast.js
aeng7: streamUrl is used for nonlinear ads and streams is used for linear ads. This is the convention that was used prior to my changes.

js/ad_manager_vast.js
aeng7: Yep, the AMF calls cancelAd.

js/freewheel.js
aeng7: Before, we were calling notifyLinearAdStarted with the name of the ad manager rather than the ad id. This is to fix that issue.

angieisrani: sounds good

test/unit-test-helpers/mock_amc.js
angieisrani: sounds good

js/ad_manager_vast.js
angieisrani: sounds good

js/ad_manager_template.js
aeng7: Added documentation to the template and this plugin.

angieisrani: could i ask you to make the tiny change of adding something to the JSDoc like this:
"... about the cancellation.  It will include the following fields:"

aeng7: Sure, will do.

js/ad_manager_vast.js
```diff
@@ -406,22 +488,48 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * Ends an ad. Notifies the AMC about the end of the ad. If it is the last linear ad in the pod,
+     * will also notify the AMC of the end of the ad pod.
+     * @public
+     * @method Vast#endAd
+     * @param {object} ad The ad to end
+     * @param {boolean} failed If true, the ending of this ad was caused by a failure
+     */
+    this.endAd = function(ad, failed) {
```
angieisrani: Since AMF is not calling endAd, why is it public?

js/ad_manager_vast.js
```diff
@@ -669,82 +775,107 @@ OO.Ads.manager(function(_, $) {
...
-     * @public
+     * @private
      * @method Vast#_handleLinearAd
+     * @param {object} ad The ad object
      * @param {object} adLoaded The ad that was loaded
-     * @returns {boolean} True if the ad was loaded and a stream was found; else false.
+     * @param {object} params Additional parameters associated with this ad
+     *                        adPodIndex : the index of the ad pod this ad is housed in
+     *                        adPodLength : the total number of ads in the ad pod this ad is housed in
+     * @returns {object} The ad unit object ready to be added to the timeline
```
angieisrani: seems to return object or boolean

angieisrani:
LGTM

PR #49, opened 2016-02-10
Pbw 4106

aeng7:
-Support for providing size of overlay to Alice
-This allows the overlay to be able to handle clickthroughs and allow the Alice UI to function alongside the overlay

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -464,7 +464,7 @@ OO.Ads.manager(function(_, $) {
           indexInPod = 0;
           if (ad.isLinear) {
             _registerDisplayForLinearAd();
-            fwContext.setParameter("renderer.video.clickDetection", false, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
+            fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_CLICK_DETECTION, false, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
             slotStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(ad) {
                 amc.notifyPodStarted(ad.id, ad.ad.getAdCount());
               }, this, ad);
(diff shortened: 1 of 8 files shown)


aeng7:
Must be merged at the same time as the PBW-4106 branch in HTML5-skin

js/freewheel.js
```diff
@@ -448,12 +460,11 @@ OO.Ads.manager(function(_, $) {
         }
         else {
           // Trigger the ad
-          _registerDisplayForNonlinearAd();
```
gfrankOoyala: Is this no longer needed?

js/freewheel.js
```diff
@@ -475,23 +492,39 @@ OO.Ads.manager(function(_, $) {
...
+              amc.notifyNonlinearAdEnded(adId);
+            }, this, ad.id);
+            delete slotStartedCallbacks[ad.ad.getCustomId()];
+            delete slotEndedCallbacks[ad.ad.getCustomId()];
           }
 
           // Register the content video wrapper to align the overlay to the correct elements
           // TODO: We also need to call this on vpaid
           // getCurrentAdInstance()._creativeRenditions[0].getCreativeApi() == "VPAID"
-          if (!ad.isLinear) {
```
gfrankOoyala: IF this is no longer needed then remove the comment.  How would this affect VPAID?

js/freewheel.js
aeng7: These are registered for their corresponding types below. If ad is linear, we register the display for a linear ad on line 466. If the ad is non linear, we register the display on line 495.

js/freewheel.js
aeng7: I think someone forgot to remove this before. The display for the nonlinear ad was registered a few lines above. There was no point to reregister it again down here since nothing has changed for the elements required for the display. I'll remove the comment.

js/freewheel.js
gfrankOoyala: Gotcha

js/google_ima.js
```diff
@@ -987,6 +1000,7 @@ require("../html5-common/js/utils/utils.js");
         if (_IMAAdsManager)
         {
           this.currentIMAAd = null;
```
pilievOoyala: These two lines with this.variable can live outside of the if check.  That way we insure they'll get cleared.

js/google_ima.js
```diff
@@ -1248,6 +1262,8 @@ require("../html5-common/js/utils/utils.js");
                 this.setVolume(this.savedVolume);
                 this.savedVolume = -1;
               }
+            } else {
```
pilievOoyala: style issue. can you space out the curly braces so it matches the rest of the file.

js/google_ima.js
aeng7: Will do.

js/google_ima.js
aeng7: Will do.

gfrankOoyala:
lgtm

mistermendez:
:+1:

PR #50, opened 2016-02-10
[PBI-744] Support error reporting events in VAST 3.0

dchauOoyala:


js/ad_manager_vast.js
```diff
@@ -64,6 +64,7 @@ OO.Ads.manager(function(_, $) {
     this.movieMd = null;
     this.adURLOverride;
     this.lastOverlayAd;
+    this.errorInfo = {};
```
gfrankOoyala: Add to comments above

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
                         companion: [],
                         linear: { tracking: {}, ClickTracking: [] },
                         nonLinear: { tracking: {} } };
+    // when wrapper ajax callback returns, wrapperParentId will be properly set
+    this.wrapperParentId = null;
```
gfrankOoyala: Please ad to comments

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * TODO: Support all error codes. Not all error events are tracked in our code.
+     * Standard VAST 3 errors
+     */
+    this.ERROR_CODES = {
```
gfrankOoyala: Please add to above comments

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+      /**
+       * TODO: Add support
+       * Video player expecting different size.
+       */
+      VIDEO_EXPECT_DIFFERENT_SIZE:        203,
+
+      /**
+       * General Wrapper Error.
+       */
+      WRAPPER:                            300,
```
gfrankOoyala: Can we rename to GENERAL_WRAPPER

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+      /**
+       * No ads VAST response after one or more Wrappers. Also includes
+       * number of empty VAST responses from fallback.
+       */
+      WRAPPER_NO_ADS:                     303,
+
+      /**
+       * General linear error. Video player is unable to display the linear ad.
+       */
+      LINEAR_ADS:                         400,
```
gfrankOoyala: GENERAL_LINEAR_ADS

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+      /**
+       * TODO: Add support
+       * Problem displaying MediaFile.
+       */
+      MEDIAFILE_DISPLAY_PROBLEM:          405,
+
+      /**
+       * General NonLinearAds error.
+       */
+      NONLINEAR_ADS:                      500,
```
gfrankOoyala: GENERAL_NONLINEAR_ADS

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+       * TODO: Add support
+       * Could not find NonLinear resource with supported type.
+       */
+      NONLINEAR_ADS_RESOURCE_UNSUPPORTED: 503,
+
+      /**
+       * TODO: Add support
+       * General CompanionAds error.
+       */
+      COMPANION_ADS:                      600,
```
gfrankOoyala: GENERAL_COMPANION_ADS

js/ad_manager_vast.js
```diff
@@ -73,6 +74,178 @@ OO.Ads.manager(function(_, $) {
...
+       * TODO: Add support
+       * Undefined error.
+       */
+      UNDEFINED:                          900,
+
+      /**
+       * TODO: Add support
+       * General VPAID error.
+       */
+      VPAID:                              901
```
gfrankOoyala: GENERAL_VPAID

gfrankOoyala: Do we support VPAID in VAST?

js/ad_manager_vast.js
dchauOoyala: Will do

js/ad_manager_vast.js
dchauOoyala: I was under the impression that, that was what Francisco was working on. Regardless, Vast 3.0 makes references to VPAID, but does not provide any specifications; it does explicitly say that there is a separate IAB guideline for it. Also, VPAID support is not part of the "Minimal Compliance" section in the VAST 3.0 specs.

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
       'close' ];
 
     /**
-     * Helper function to verify that the xml has a valid vast ad in it and that it is a valid xml.
+     * Helper function to verify that XML is valid
      * @param {xml} vastXML Contains the vast ad data to be parsed.
```
gfrankOoyala: can a @public tot he comments if making this public or @private if not, why is this public?

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
      * @returns {boolean} Returns true if the xml is valid otherwise it returns false.
      */
-    var isValidVastXML = _.bind(function(vastXML) {
+    this.isValidVastXML = _.bind(function(vastXML) {
```
gfrankOoyala: if making this public you don't need to _.bind it any more

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
+      return this.isValidRootTagName(vastXML) && this.isValidVastVersion(vastXML);
+    }, this);
+
+    /**
+     * Helper function to verify XML has valid VAST root tag.
+     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * @returns {boolean} Returns true if the root tag is valid otherwise it returns false.
```
gfrankOoyala: add an @public to comments. Why is this public?

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
+        this.trigger(this.ERROR, this);
+        return false;
+      }
+      return true;
+    }, this);
+
+    /**
+     * Helper function to verify XML is a valid VAST version.
+     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * @returns {boolean} Returns true if the VAST version is valid otherwise it returns false.
```
gfrankOoyala: Add an @public why is this public?

js/ad_manager_vast.js
```diff
@@ -541,6 +785,52 @@ OO.Ads.manager(function(_, $) {
...
     /**
+    * This method pings all the ad's error URLs with a specific error code if the error URL
+    * contains the macro, "[ERRORCODE]". If ad has a parent wrapper, then go up the chain and ping
+    * wrapper's error urls as well.
+    * @public
+    * @method Vast#trackError
+    * @param {number} code Error code.
+    * @param {boolean} currentAdId Ad ID of current ad.
+    */
+    this.trackError = _.bind(function trackErrorHelper(code, currentAdId) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
+     *
+     * <Error> tags are also optional so they may not always exist.
+     * @public
+     * @method Vast#getErrorInfo
+     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * @param {string} wrapperParentIdArg Is the current ad's "parent" wrapper ID. Could be
+     * undefined if ad does not have parent/wrapper.
+     * @returns {object} The error object with a list of error urls and whether or not there are no ads
+     */
+    this.getErrorInfo = _.bind(function(vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
+      }
+      return true;
+    }, this);
+
+    /**
+     * Helper function to verify XML is a valid VAST version.
+     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * @returns {boolean} Returns true if the VAST version is valid otherwise it returns false.
+     */
+    this.isValidVastVersion = _.bind(function(vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -83,22 +256,89 @@ OO.Ads.manager(function(_, $) {
...
+    this.isValidVastXML = _.bind(function(vastXML) {
+      return this.isValidRootTagName(vastXML) && this.isValidVastVersion(vastXML);
+    }, this);
+
+    /**
+     * Helper function to verify XML has valid VAST root tag.
+     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * @returns {boolean} Returns true if the root tag is valid otherwise it returns false.
+     */
+    this.isValidRootTagName = _.bind(function(vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -872,68 +1212,104 @@ OO.Ads.manager(function(_, $) {
...
+    /**
+     * Helper function to handle Inline Ad
+     * @public
+     * @method Vast#handleInline
+     * @param {object} adLoaded The ad loaded object and metadata.
+     * @param {object} vastAd The object returned from parser, containing ad information (tracking events, linear/nonlinear
+     * ad information, companion ads, etc)
+     * @param {xml} vastXML The current vast xml that contains the ad data.
+     */
+    this.handleInline = _.bind(function(adLoaded, vastAd, vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -872,68 +1212,104 @@ OO.Ads.manager(function(_, $) {
      * @method Vast#_onVastResponse
      * @param {object} adLoaded The ad loaded object and metadata.
      * @param {object} xml The xml returned from loading the ad.
+     * @param {string} wrapperParentIdArg Is the current ad's "parent" wrapper ID. This argument would be set on an ajax
+     * call for a wrapper ad. This argument could also be undefined if ad did not have parent/wrapper.
      */
-    this._onVastResponse = function(adLoaded, xml) {
+    this._onVastResponse = _.bind(function(adLoaded, xml, wrapperParentIdArg) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -575,51 +865,98 @@ OO.Ads.manager(function(_, $) {
...
+      }
+      return false;
+    }, this);
+
+    /**
      * If a non-linear Ad is found then it is parsed and added to the timeline via the addToTimeline function.
      * @public
      * @method Vast#_handleNonLinearAd
      * @param {object} adLoaded The ad that was loaded.
+     *
```
gfrankOoyala: Space not needed

js/ad_manager_vast.js
```diff
@@ -575,51 +865,98 @@ OO.Ads.manager(function(_, $) {
...
     };
 
     /**
+     * Helper function to determine whether a nonlinear ad exists
+     * @public
+     * @method Vast#hasNonLinear
+     * @param {xml} vastXML The current vast xml that contains the ad data.
+     * @returns {boolean} True if vastXML has an Inline NonLinear ad, False otherwise.
+     */
+    this.hasNonLinear = _.bind(function(vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -872,68 +1212,104 @@ OO.Ads.manager(function(_, $) {
...
+    /**
+     * Helper function to handle Wrapper Ad
+     * @public
+     * @method Vast#handleWrapper
+     * @param {object} adLoaded The ad loaded object and metadata.
+     * @param {object} vastAd The object returned from parser, containing ad information (tracking events, linear/nonlinear
+     * ad information, companion ads, etc)
+     * @param {xml} vastXML The current vast xml that contains the ad data.
+     */
+    this.handleWrapper = _.bind(function(adLoaded, vastAd, vastXML) {
```
gfrankOoyala: no need to bind if it is public

js/ad_manager_vast.js
```diff
@@ -94,19 +269,43 @@ OO.Ads.manager(function(_, $) {
...
-     * Helper function to verify that the xml has a valid vast ad in it and that it is a valid xml.
-     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * Helper function to verify that XML is valid
+     * @param {xml} vastXML Contains the vast ad data to be parsed
      * @returns {boolean} Returns true if the xml is valid otherwise it returns false.
      */
-    var isValidVastXML = _.bind(function(vastXML) {
+    this.isValidVastXML = _.bind(function(vastXML) {
+      return this.isValidRootTagName(vastXML) && this.isValidVastVersion(vastXML);
+    }, this);
```
angieisrani: since this is public it not longer needs to be bound

angieisrani: same with those below

js/ad_manager_vast.js
```diff
@@ -94,19 +269,43 @@ OO.Ads.manager(function(_, $) {
       'close' ];
 
     /**
-     * Helper function to verify that the xml has a valid vast ad in it and that it is a valid xml.
-     * @param {xml} vastXML Contains the vast ad data to be parsed.
+     * Helper function to verify that XML is valid
```
angieisrani: can you add the method name to the jsdoc?
@method Class#methodname

js/ad_manager_vast.js
```diff
@@ -94,19 +269,43 @@ OO.Ads.manager(function(_, $) {
...
+     * Helper function to verify XML has valid VAST root tag.
+     * @param {xml} vastXML Contains the vast ad data to be parsed
+     * @returns {boolean} Returns true if the root tag is valid otherwise it returns false.
+     */
+    this.isValidRootTagName = _.bind(function(vastXML) {
       var rootTagName = (vastXML && vastXML.firstChild) ? vastXML.firstChild.tagName || '' : '';
       if (rootTagName.toUpperCase() != "VAST") {
         OO.log("Invalid VAST XML for tag name: " + rootTagName);
+        this.trackError(this.ERROR_CODES.SCHEMA_VALIDATION, this.wrapperParentId);
+        this.errorType = "schemaValidationError";
```
angieisrani: instead of hardcoding this, can you associate the this.ERROR_CODES object with both the name and number?
something like:
this.ERROR_CODES = { SCHEMA_VALIDATION: { id: 101, type: "schemaValidationError" }, ...}

angieisrani: It looks like errorType is always followed by a call to this.trigger.  What is this.errorType for?  Can it just be passed as a parameter?

angieisrani: Same with the trackError versus trigger functions.  Looks like you always do these three things in a row.  Can you wrap this in a single function?

js/ad_manager_vast.js
```diff
@@ -150,14 +351,53 @@ OO.Ads.manager(function(_, $) {
...
 
     /**
+     * Helper function to grab error information. VastAdSingleParser already grabs error data while
+     * creating ad object, but some errors may occur before the object is created.
+     * Note: <Error> can only live in three places: directly under <VAST>, <Ad>, or <Wrapper> elements.
+     * <Error> tags are also optional so they may not always exist.
+     * @public
+     * @method Vast#getErrorInfo
+     * @param {xml} vastXML Contains the vast ad data to be parsed
+     * @param {string} wrapperParentIdArg Is the current ad's "parent" wrapper ID. Could be
```
angieisrani: i don't see this param below?

js/ad_manager_vast.js
```diff
@@ -150,14 +351,53 @@ OO.Ads.manager(function(_, $) {
...
+     * undefined if ad does not have parent/wrapper.
+     * @returns {object} The error object with a list of error urls and whether or not there are no ads.
+     */
+    this.getErrorInfo = _.bind(function(vastXML) {
```
angieisrani: is this looking for errors of getting the error urls?  I'm a little confused whether this is doing validation or parsing.  Looks like mainly just parsing data.  Possibly consider renaming to ErrorTrackingInfo?

js/ad_manager_vast.js
```diff
@@ -640,6 +884,52 @@ OO.Ads.manager(function(_, $) {
...
+    * @param {boolean} currentAdId Ad ID of current ad
+    */
+    this.trackError = _.bind(function trackErrorHelper(code, currentAdId) {
+      if (currentAdId && currentAdId in this.errorInfo) {
+        pingURLs(this.errorInfo[currentAdId].errorUrls);
+        var parentId = this.errorInfo[currentAdId].wrapperParentId;
+
+        // ping parent wrapper's error urls too if ad had parent
+        if (parentId) {
+          this.trackErrorHelper(code, parentId);
```
angieisrani: Hm - does the function need two names?  Does trackError not map exactly to trackErrorHelper except that the inner one is not bound to this?

js/ad_manager_vast.js
```diff
@@ -640,6 +884,52 @@ OO.Ads.manager(function(_, $) {
...
+      url = url.replace(/\[ERRORCODE\]/, code);
+      OO.pixelPings(url);
+    }, this);
+
+    /**
+     * Helper function to ping error URLs.
+     * @private
+     * @method Vast#pingURL
+     * @param {code} code Error code
+     * @param {string} urls URLs to ping
```
angieisrani: should be type {object} with description saying it's an arrray

js/ad_manager_vast.js
```diff
@@ -640,6 +884,52 @@ OO.Ads.manager(function(_, $) {
...
+    var pingURL = _.bind(function(code, url) {
+      url = url.replace(/\[ERRORCODE\]/, code);
+      OO.pixelPings(url);
+    }, this);
+
+    /**
+     * Helper function to ping error URLs.
+     * @private
+     * @method Vast#pingURL
+     * @param {code} code Error code
```
angieisrani: code is not the type - is it number?

js/ad_manager_vast.js
```diff
@@ -672,53 +962,101 @@ OO.Ads.manager(function(_, $) {
...
-        // No Playable stream, report error.
-        OO.log("Can not find playable stream in vast result", this.inlineAd);
-        return false;
+    this._handleLinearAd = function(adLoaded, vastXML) {
+      // See if the <Linear> even exists
+      if (this.hasLinear(vastXML)) {
+        // filter our playable stream:
+        var firstLinearAd = _.find(this.inlineAd.ads, function(v){ return !_.isEmpty(v.linear.mediaFiles); }, this);
+        if (!firstLinearAd) {
+          OO.log("General Linear Ads Error: no mediafiles", this.inlineAd);
```
angieisrani: if possible, prefix logs with "Vast"

js/ad_manager_vast.js
dchauOoyala: sorry, that was sloppy of me. I forgot to remove it from the doc

js/ad_manager_vast.js
```diff
@@ -672,53 +962,101 @@ OO.Ads.manager(function(_, $) {
...
+     * @param {xml} vastXML The current vast xml that contains the ad data
+     * @returns {boolean} true if vastXML has an Inline Linear ad, false otherwise.
+     */
+    this.hasLinear = _.bind(function(vastXML) {
+      var inlineElement = $(vastXML).find("InLine");
+      var linearElement = $(inlineElement).find("Linear");
+      if (linearElement.length > 0) {
+        return true;
+      }
+      return false;
```
angieisrani: consider:
```
return (linearElement.length > 0) ;
```

js/ad_manager_vast.js
```diff
@@ -672,53 +962,101 @@ OO.Ads.manager(function(_, $) {
...
+     * @param {xml} vastXML The current vast xml that contains the ad data
+     * @returns {boolean} true if vastXML has an Inline NonLinear ad, false otherwise.
+     */
+    this.hasNonLinear = _.bind(function(vastXML) {
+      var inlineElement = $(vastXML).find("InLine");
+      var nonLinearElement = $(inlineElement).find("NonLinear");
+      if (nonLinearElement.length > 0) {
+        return true;
+      }
+      return false;
```
angieisrani: consider 
```
return (nonLinearElement.length > 0);
```

js/ad_manager_vast.js
```diff
@@ -776,10 +1114,10 @@ OO.Ads.manager(function(_, $) {
...
      * @private
      * @method Vast#parseTrackingEvents
-     * @param {array} tracking to add the tracking info to and return.
-     * @param {xml} xml The data of the ad with tracking info.
-     * @param {array} trackingEvents List of events that are tracked, if null then it uses the global one.
-     * @returns {array} tracking An array of tracking items.
+     * @param {array} tracking to add the tracking info to and return
+     * @param {xml} xml The data of the ad with tracking info
+     * @param {array} trackingEvents List of events that are tracked, if null then it uses the global one
+     * @returns {array} An array of tracking items.
```
angieisrani: i thought "array" is not a supported type by our JSDoc interpreter?  I thought it had to be object?

js/ad_manager_vast.js
```diff
@@ -793,7 +1131,8 @@ OO.Ads.manager(function(_, $) {
      * Helper function that make sure the array is not empty.
```
angieisrani: removes empty items?

js/ad_manager_vast.js
```diff
@@ -793,7 +1131,8 @@ OO.Ads.manager(function(_, $) {
...
      * @private
      * @method Vast#filterEmpty
-     * @param {array} array An array that is the be checked if it is empty.
+     * @param {array} array An array that is the be checked if it is empty
+     * @returns {array} The filtered array.
      */
     var filterEmpty = _.bind(function(array) {
       return _.reject(array, function(x){
```
angieisrani: consider _.without
http://underscorejs.org/#without

js/ad_manager_vast.js
```diff
@@ -837,8 +1176,8 @@ OO.Ads.manager(function(_, $) {
      * The xml needs to be parsed to grab all the linear data of the ad and create an object.
      * @private
      * @method Vast#parseLinearAd
-     * @param {xml} Xml containing the ad data to be parsed.
-     * @returns {object} result An object containing the ad data.
+     * @param {xml} linearXml The xml containing the ad data to be parsed
```
angieisrani: xml is a recognized JS object type?  Should this be string?

angieisrani: this is seen in many places

js/ad_manager_vast.js
dchauOoyala: You're right. Just realized I inadvertently made this function do both the validation and parsing. It was meant to just do the parsing, but while writing it I found an opportunity to catch an error here. I should place this GENERAL_WRAPPER_NO_ADS error check in a separate function independent from this one.

js/ad_manager_vast.js
```diff
@@ -924,6 +1263,7 @@ OO.Ads.manager(function(_, $) {
       var nonLinearAds = $(xml).find("NonLinearAds");
 
       if (type === "wrapper") { result.VASTAdTagURI = $(xml).find("VASTAdTagURI").text(); }
+      result.id = $(xml).prop("id");
       result.error = filterEmpty($(xml).find("Error").map(function() { return $(this).text(); }));
```
angieisrani: consider saving $(xml) to a variable to avoid the repeated conversion

js/ad_manager_vast.js
dchauOoyala: Will do

js/ad_manager_vast.js
angieisrani: That's fine - it's up to you where it should be.  I think the name is what confused me.  It looked like I would call this to find out what errors were found but the code looks like it's actually just parsing tracking urls

js/ad_manager_vast.js
dchauOoyala: yes, that is more concise. will do

js/ad_manager_vast.js
dchauOoyala: you're right, there isn't an xml primitive. I'll make the change for the rest of them

js/ad_manager_vast.js
dchauOoyala: I will revise this sentence

js/ad_manager_vast.js
dchauOoyala: Will do

js/ad_manager_vast.js
dchauOoyala: It is most definitely number

js/ad_manager_vast.js
dchauOoyala: got it, will change

js/ad_manager_vast.js
dchauOoyala: Yeah "array" definitely isn't supported. You can actually just add "[]" to a type and that will connote that it is an array. So in this case, "tracking" is an array of strings so the syntax should be "{string[]}". I'll make this change for the rest of them. Here's the docs @stephanieblotner pointed me to: http://usejsdoc.org/tags-type.html

angieisrani: That's awesome - from my testing, though, i found that our interpretter didn't exactly match that doc and i couldn't get this array thing working when i last tried it.  @stephanieblotner are you able to run this file through your tool to see what comes out here?

test/unit-tests/vast_test.js
```diff
@@ -584,4 +591,202 @@ describe('ad_manager_vast', function() {
...
+    var content = {
+      embed_code: embed_code,
+      ads: [vast_ad_mid]
+    };
+    vastAdManager.initialize(amc);
+    vastAdManager.currentDepth = OO.playerParams.maxVastWrapperDepth;
+    vastAdManager._onVastResponse(vast_ad_mid, wrapperXML);
+    expect(vastAdManager.errorType).to.be("tooManyWrapper");
+    vastAdManager.errorType = '';
+  });
```
aeng7: Is the errorType only used for unit tests? The original jsdoc mentioned its used for logging but I do not observe it being used in the code. If it is no longer necessary in the base code, it might be a good idea to remove it.

aeng7: The unit tests could be made to be a little more robust by checking against the external APIs it will be working with, rather than checking against a property. Do all the errors eventually call OO.pixelPings? If so, you can mock pixelPings to check if it is called properly for these unit tests.

dchauOoyala: got it. will do

js/ad_manager_vast.js
dchauOoyala: Will consider this. Also, after talking with alex, it seems like errorType and this.trigger() are not used anymore.

js/ad_manager_vast.js
dchauOoyala: Hmm, so I was initially advised to use a named function expression for recursive functions. Not entirely understand why, I found this on stackoverflow (http://stackoverflow.com/questions/7065120/calling-a-javascript-function-recursively). So I did it mainly for this, but since we don't really have that issue in our code, actually it doesn't make much sense to use a named function expression. I can remove it!

angieisrani: I see - i think the name only matters if you plan on assigning it to another function and changing it.  Since this function won't change or be mapped to another function name i think it's okay to stay with one name

dchauOoyala: got it.

js/ad_manager_vast.js
dchauOoyala: actually, i did a quick sanity check and called typeof on the vastXML that comes back. It's actually of object type (XMLDocument to be specific). I think it's appropriate to just use "object", then.

js/ad_manager_vast.js
```diff
@@ -150,14 +351,66 @@ OO.Ads.manager(function(_, $) {
...
     /**
+     * Helper function to grab error information. VastAdSingleParser already grabs error data while
+     * creating ad object, but some errors may occur before the object is created.
+     * Note: <Error> can only live in three places: directly under <VAST>, <Ad>, or <Wrapper> elements.
+     * <Error> tags are also optional so they may not always exist.
+     * @public
+     * @method Vast#getErrorInfo
+     * @param {XMLDocument} vastXML Contains the vast ad data to be parsed
+     * @param {object} ads A jQuery object which contains the collection of ad elements found
+     * @returns {object} The error object with a list of error urls and whether or not there are no ads.
```
angieisrani: I don't see a return below?

js/ad_manager_vast.js
```diff
@@ -790,23 +1141,22 @@ OO.Ads.manager(function(_, $) {
     }, this);
 
     /**
-     * Helper function that make sure the array is not empty.
+     * Helper function to remove empty items.
      * @private
      * @method Vast#filterEmpty
-     * @param {array} array An array that is the be checked if it is empty.
+     * @param {Array} array An array that is the be checked if it is empty
+     * @returns {Array} The filtered array.
```
angieisrani: just verifying -is Array a type?

dchauOoyala: So according to: https://msdn.microsoft.com/en-us/library/7wkd9z69(v=vs.94).aspx
Array is a "composite" type and falls under the same category as Object, so I want to say that it is. 

Also, I did a quick double check on stackoverflow (because the JSDoc docs don't have many Array cases): http://stackoverflow.com/questions/14611995/how-to-specify-an-array-of-objects-as-a-parameter-or-return-value-in-jsdoc
We are using JSDoc3 so the use of Array does seem kosher.

angieisrani: sounds good

aeng7:
LGTM

PR #52, opened 2016-02-18
[PBW-4106]

aeng7:
IMA:
-Fixed an issue where overlay would get cancelled preemptively


gfrankOoyala:
Not enough code changes, denied. Just kidding. Love 1 line code fixes. So easy to CR. LGTM but how about a unit test

aeng7:
Will modify a current one.

aeng7:
Unit test in.

gfrankOoyala:
lgtm

PR #53, opened 2016-02-18
[PBI-731a]

dchauOoyala:
Unsubscribe to fw_adRequestComplete to properly timeout ad request

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -285,9 +285,10 @@ OO.Ads.manager(function(_, $) {
     var _adRequestTimeout = _.bind(function(){
       _clearAdRequestTimeout();
       if (!fwContext._adResponse) {
-        var error = "ad request timeout";
-        fw_onError(error);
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -304,7 +305,7 @@ OO.Ads.manager(function(_, $) {
     var _setAdRequestTimeout = _.bind(function(callback, duration){
       if (adRequestTimeout) {
         var error = "Ad Request Timeout already exists - bad state";
-        fw_onError(error);
+        fw_onError(null, error);
```
gfrankOoyala: Why null? What is the first param?

dchauOoyala: So the fw_onError function on line 816 takes in an event object and a string error message. The function doesn't actually do anything with the event, so I passed in null. Should I change the original function?

gfrankOoyala: if it is possible to change the original without breaking it, yes, otherwise no. Because if we don't use that param and it isn't required (aka sdk required)  then we should remove the param so it is less confusing.

gfrankOoyala:
LGTM

PR #54, opened 2016-02-18
[PBI-1454] Fix error reporting issue and cleanup

dchauOoyala:
- Add function, _hasLinearAd() to check if the ad has a linear ad before calling _handleLinearAd()
- Add function, _hasNonLinearAd() to check if the ad has a linear ad before calling _handleNonLinearAd()
- Code Cleanup and JSDoc fix

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -664,7 +664,7 @@ OO.Ads.manager(function(_, $) {
         }, this);
         this.checkCompanionAds(adWrapper.ad);
         initSkipAdOffset(adWrapper);
-        var hasClickUrl = adWrapper.ad.data.linear.ClickThrough.length > 0;
+        var hasClickUrl = adWrapper.ad.data.linear.clickThrough.length > 0;
         this.amc.notifyLinearAdStarted(adWrapper.id, {
           name: adWrapper.ad.data.title,
           duration: adWrapper.ad.durationInMilliseconds/1000,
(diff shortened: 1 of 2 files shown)


js/ad_manager_vast.js
```diff
@@ -1067,6 +1067,40 @@ OO.Ads.manager(function(_, $) {
...
 
     /**
+     * Helper function to determine if the ad is a linear ad.
+     * @private
+     * @method Vast#_hasLinearAd
+     * @param {object} ad The ad object
+     * @returns {boolean} true if the ad is a linear ad, false otherwise.
+     */
+    var _hasLinearAd = _.bind(function(ad) {
+      var linearObjectKeys = _.keys(ad.linear);
```
gfrankOoyala: what if there is no clickthrough? Should we check the creative instead?

js/ad_manager_vast.js
```diff
@@ -1067,6 +1067,40 @@ OO.Ads.manager(function(_, $) {
...
+     * @returns {boolean} true if the ad is a nonlinear ad, false otherwise.
+     */
+    var _hasNonLinearAd = _.bind(function(ad) {
+      var nonLinearObjectKeys = _.keys(ad.nonLinear);
+      // Magic number 1 because after mergeVastAdResult() is called, the ad object
+      // has will have an object associated to the key, "nonLinear". This object always
+      // has at least 1 key (tracking) even if a nonlinear creative does not exist. If one
+      // does exist, there must be more than 1 key; there must be a "url" key as specified
+      // by the Vast 3.0 specifications.
+      return (nonLinearObjectKeys.length > 1);
```
gfrankOoyala: What if there is a clickthrough?

gfrankOoyala:
Why not use the isLinear functions?

js/ad_manager_vast.js
dchauOoyala: so the way it originally was in our code, clickTracking and tracking will always be available in the linear object. It's not great, so I'm in the middle of changing it so I don't need to use these magic values of 1/2

dchauOoyala:
The isLinear properties are only available when the ad is played. I do not think the isLinear property is available to whatever "ad" object is in context when we call handleLinearAd/handleNonLinearAd

gfrankOoyala:
Sure but instead of checking clickthrough and tracking why not check for creative instead? Seems like assuming 2 or 1 is a magic number is not accurate. What if there is no clickthrough or error tracking urls but a creative none the less?

dchauOoyala:
I see what you're saying, that's actually the way I originally made hasLinearAd/hasNonLinearAd work; check the vastXML to see if the creative exists. But, that got lost in the midst of merging with the podded ads code. I can do it that way instead of checking against the ad object.

aeng7:
LGTM

PR #55, opened 2016-02-19
[PBI-1459]

aeng7:
-fixed an issue where FW video ads would not play on mobile devices
-However, this also makes it overlay ads cannot be interacted with on iOS devices (mainly will affect iPad). We'll have to live with this for now until Freewheel provides us a way to do so

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -248,10 +248,18 @@ OO.Ads.manager(function(_, $) {
       fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_DISPLAY_CONTROLS_WHEN_PAUSE, false, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
       fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_CLICK_DETECTION, true, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
 
-      // position the overlay y index to 0. Alice will take care of positioning the overlay above the control bar
-      // NOTE: If we set renderer.html.coadScriptName we can probably render overlays on our own
-      //       (coad stands for customer owned ad renderer)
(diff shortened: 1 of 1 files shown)


js/freewheel.js
```diff
@@ -962,6 +977,16 @@ OO.Ads.manager(function(_, $) {
     }, this);
 
     /**
+     * Checks to see if we can use a fake video to trick Freewheel into showing overlays
+     * in a div that is intregrated with Alice.
+     * @private
+     * @method Freewheel#_supportsSecondaryVideoForOverlay
+     */
+    var _supportsSecondaryVideoForOverlay = _.bind(function() {
```
angieisrani: There is an API in environment.js that you may be able to use here

gfrankOoyala: Is it necessary to put this into a function? Just wondering, since it is technically a function calling a function.

js/freewheel.js
```diff
@@ -248,10 +248,18 @@ OO.Ads.manager(function(_, $) {
...
       fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_CLICK_DETECTION, true, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
 
-      // position the overlay y index to 0. Alice will take care of positioning the overlay above the control bar
-      // NOTE: If we set renderer.html.coadScriptName we can probably render overlays on our own
-      //       (coad stands for customer owned ad renderer)
-      fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_HTML_MARGIN_HEIGHT, 0, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
+      if (OO.requiresSingleVideoElement) {
+        // NOTE: If we set renderer.html.coadScriptName we can probably render overlays on our own
+        //       (coad stands for customer owned ad renderer)
+        var controlHeight = amc.ui.rootElement.find(".controlBar");
```
angieisrani: do we get this from an event from alice?

angieisrani:
LGTM

PR #56, opened 2016-02-20
[PBW-4447][VTC] Pass asset's delivery type on setting video plugin's URL

mityaha:
* Added encoding parameter to setVideoUrl

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1843,10 +1843,11 @@ require("../html5-common/js/utils/utils.js");
      * Sets the url of the video.
      * @public
      * @method GoogleIMAVideoWrapper#setVideoUrl
+     * @param {string} encoding The encoding of video stream 
      * @param {string} url The new url to insert into the video element's src attribute
      * @returns {boolean} True or false indicating success
      */
-    this.setVideoUrl = function(url)
+    this.setVideoUrl = function(encoding, url)
     {
       return true;
     };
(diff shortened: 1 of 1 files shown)


js/google_ima.js
```diff
@@ -1843,10 +1843,11 @@ require("../html5-common/js/utils/utils.js");
      * Sets the url of the video.
      * @public
      * @method GoogleIMAVideoWrapper#setVideoUrl
+     * @param {string} encoding The encoding of video stream 
      * @param {string} url The new url to insert into the video element's src attribute
      * @returns {boolean} True or false indicating success
      */
-    this.setVideoUrl = function(url)
+    this.setVideoUrl = function(encoding, url)
```
angieisrani: I would advise making this the second parameter to avoid breaking other plugins

js/google_ima.js
```diff
@@ -1844,9 +1844,10 @@ require("../html5-common/js/utils/utils.js");
      * @public
      * @method GoogleIMAVideoWrapper#setVideoUrl
      * @param {string} url The new url to insert into the video element's src attribute
+     * @param {string} encoding The encoding of video stream
```
angieisrani: This is fine; since you changed the comment in other places to include OO.VIDEO.ENCODING it could be done here as well but is not important.  Although actually since this field isn't used by IMA you don't need to include this at all; it should still work without this change.  Especially since IMA won't be able to use the encoding field based on its implementation

mityaha: Done, removed changes

angieisrani:
I see that in the end there are no changes to merge with this; Do you mind if we close the pull request instead of merging?

mityaha:
Yes, I will delete it

PR #57, opened 2016-02-22
[PBI-1473] Change nonLinearClickThrough to NonLinearClickThrough

dchauOoyala:


aeng7:
LGTM. There's an issue with the unit test that would have caught this. Douglas is looking into that.

PR #59, opened 2016-02-26
[PBI-1479]

aeng7:
IMA
-IMA will now request the AMC to hide the player UI so that IMA UI elements are not blocked


dchauOoyala:
LGTM

aeng7:
Needs to be merged the same time as mjolnir PR 521

PR #60, opened 2016-03-01
[PBI-799] Add initial VMAP functionality

dchauOoyala:


js/ad_manager_vast.js
```diff
@@ -1616,6 +1645,218 @@ OO.Ads.manager(function(_, $) {
...
+         *ad_set_code: "",
+         *click_url: "",
+         *expires: 0,
+         *first_shown: 0,
+         *frequency: 2,
+         *public_id: "",
+         *signature: "",
+         *tracking_url: [],
+         *type: "",
+         *url: ""
```
dchauOoyala: Are the rest of these properties important? I only use the "time" and "position_type" properties.

js/ad_manager_vast.js
```diff
@@ -282,9 +283,8 @@ OO.Ads.manager(function(_, $) {
      * @returns {boolean} Returns true if the root tag is valid otherwise it returns false.
      */
     this.isValidRootTagName = function(vastXML) {
-      var rootTagName = (vastXML && vastXML.firstChild) ? vastXML.firstChild.tagName || '' : '';
-      if (rootTagName.toUpperCase() != "VAST") {
-        OO.log("VAST: Invalid VAST XML for Tag Name: " + rootTagName);
+      if ($(vastXML).find("VAST").length === 0) {
+        OO.log("VAST: Invalid VAST XML");
```
aeng7: Is it possible for a the XML to have "VAST" elsewhere in the XML and not in the root tag, causing this function to return  true when it should be false because it will be found by "$(vastXML).find("VAST").length"

js/ad_manager_vast.js
```diff
@@ -316,7 +316,8 @@ OO.Ads.manager(function(_, $) {
      * @returns {string} The Vast version.
      */
     var getVastVersion = _.bind(function(vastXML) {
-      return $(vastXML.firstChild).attr('version');
+      var vastTag = $(vastXML).find("VAST")[0];
```
aeng7: Related to the comment for isValidRootTagName, maybe we can introduce a getVastRoot function to retrieve the root of the Vast Tag and use it when we need to reference the root, like for validation. What do you think?

js/ad_manager_vast.js
aeng7: Where did you get these properties from?

js/ad_manager_vast.js
```diff
@@ -1616,6 +1645,218 @@ OO.Ads.manager(function(_, $) {
...
+            break;
+          // case: hh:mm:ss.mmm | hh:mm:ss
+          case /^\d{2}:\d{2}:\d{2}\.000$|^\d{2}:\d{2}:\d{2}$/.test(adBreak.timeOffset):
+            adObject.position_type = "t";
+            adObject.time = _convertTimeStampToSeconds(adBreak.timeOffset);
+            break;
+          // case: [0, 100]%
+          case /^\d{1,3}%$/.test(adBreak.timeOffset):
+            // TODO: test percentage > 100
+            adObject.position_type = "t";
```
aeng7: Can position_type be something other than t in this function? We can pull it out of each case statement and place it above the switch statement if so.

js/ad_manager_vast.js
```diff
@@ -1616,6 +1645,218 @@ OO.Ads.manager(function(_, $) {
...
+     * @method Vast#_parseAdBreak
+     * @param {object} adBreakElement The adBreak element to parse
+     * @returns {object} The formatted adBreak object.
+     */
+    var _parseAdBreak = _.bind(function(adBreakElement) {
+      var adBreak = {};
+      adBreak.timeOffset = $(adBreakElement).attr("timeOffset");
+      adBreak.breakType = $(adBreakElement).attr("breakType");
+      adBreak.breakId = $(adBreakElement).attr("breakId");
+      return adBreak;
```
aeng7: Do we not want to handle repeatAfter, as per VMAP 1.0.1 spec?

js/ad_manager_vast.js
dchauOoyala: "adObject" is analogous to the "adLoaded" object that is passed into onVastResponse(). I got the properties by throwing in a breakpoint and seeing what properties "adLoaded" had.

js/ad_manager_vast.js
dchauOoyala: Hmm, it's not possible for there to be another VAST tag anywhere else in a VAST 3.0 response (as specified by the VAST 3 specs) so if there is another one, then it would be an error in the XML.

js/ad_manager_vast.js
dchauOoyala: sounds like a good idea

js/ad_manager_vast.js
dchauOoyala: Yes we do want to support repeatAfter; I'll add a ticket for that support.

js/ad_manager_vast.js
dchauOoyala: So according to our docs (https://wiki.corp.ooyala.com/display/idocs/Player+V4+IMA+Review+Draft#link-talk-1417), position_type "p" for percentages should also be acceptable, but when I tried testing it with "p" and passing the appropriate percentage, it didn't seem to work, but yeah we can pull it out of the switch if only is "t" is available. I'll double check

js/ad_manager_vast.js
aeng7: Got it. Can we use the new getVastRoot function to replace "$(vastXML).find("VAST")"?

js/ad_manager_vast.js
aeng7: Awesome, thanks!

js/ad_manager_vast.js
dchauOoyala: Double checked. Tried using position_type "p" like in the IMA docs and using the position property instead; does not properly hmm. I also have not been able to find where the VAST properties originate from. But yes, I will pull the position_type statements out of the switch.

js/ad_manager_vast.js
aeng7: Got it. If everything works, I think we're ok in not having these defined since you only use the two properties time and position_type.

aeng7:
LGTM!

PR #61, opened 2016-03-02
[PBI-1503]

aeng7:
IMA
-added VPAID support
-removed usage of AUTO_SCALE since IMA VPAID does not currently seem to support it. We will grab the size from the uiContainer
-fixed an issue where IMA expected the AD_LOAD_TIMEOUT setting to be in milliseconds rather than seconds


aeng7:
Needs to be merged same time as Mjolnir PR 526

pilievOoyala:
LGTM. Will wait on merge till pr 526

PR #62, opened 2016-03-04
[PBI-1484]

aeng7:
-fixed an issue where it was possible to trigger a window.open leading to a blank page with a Vast ad without a clickthrough defined
-added a new mock xml to test for when there is no clickthrough defined for a linear ad

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -1299,7 +1299,8 @@ OO.Ads.manager(function(_, $) {
         tracking: {},
         // clickTracking needs to be remembered because it can exist in wrapper ads
         clickTracking: filterEmpty($(linearXml).find("ClickTracking").map(function() { return $(this).text(); })),
-        clickThrough: filterEmpty($(linearXml).find("ClickThrough").map(function() { return $(this).text(); })),
+        //There can only be one clickthrough as per Vast 2.0/3.0 specs and XSDs
+        clickThrough: $(linearXml).find("ClickThrough").text(),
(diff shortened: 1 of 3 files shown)


aeng7:
Changed detection of clickthrough to only consider a single element. From the Vast 3.0 XSD:

<xs:element name="ClickThrough" minOccurs="0" maxOccurs="1">

js/ad_manager_vast.js
```diff
@@ -1299,7 +1299,8 @@ OO.Ads.manager(function(_, $) {
         tracking: {},
         // clickTracking needs to be remembered because it can exist in wrapper ads
         clickTracking: filterEmpty($(linearXml).find("ClickTracking").map(function() { return $(this).text(); })),
-        clickThrough: filterEmpty($(linearXml).find("ClickThrough").map(function() { return $(this).text(); })),
+        //There can only be one clickthrough as per Vast 2.0/3.0 specs and XSDs
+        clickThrough: $(linearXml).find("ClickThrough").text(),
```
dchauOoyala: Does this return undefined/null if there aren't any ClickThrough's?

dchauOoyala:
LGTM

js/ad_manager_vast.js
aeng7: find() returns a jQuery object which contains all the found elements. If there is nothing found, the jQuery object will still be returned, but it will contain no elements.

dchauOoyala: Ah, sorry I meant to ask, if the .text() will return undefined/null

PR #68, opened 2016-03-18
VAST/VPAID refactor

nevermind89x:
Still missing some stuff to refactor, but I feel at this point this was the safest changes I was able to make. This should also fix some (if not all) of the issues with those fake podded ads from backlot. Hopefully won't cause any others.


js/ad_manager_vast.js
```diff
@@ -483,6 +546,85 @@ OO.Ads.manager(function(_, $) {
...
       this.amc.addPlayerListener(this.amc.EVENTS.INITIAL_PLAY_REQUESTED, _.bind(this.initialPlay, this));
+      this.amc.addPlayerListener(this.amc.EVENTS.FULLSCREEN_CHANGED, _.bind(_onFullscreenChanged, this));
+      this.amc.addPlayerListener(this.amc.EVENTS.SIZE_CHANGED, _onSizeChanged);
+    };
+
+    /**
+     * Callback when the media file is loaded. Once is loaded we can initialize the ad
+     * @private
+     */
+    var _initializeAd = _.bind(function() {
```
aeng7: If this function is used for VPAID only, let's make note of this in the jsdoc and function name. With the influx of new VPAID functions, we should try to make these distinctions where appropriate. Maybe even encapsulate them into a separate object? Though that would be a talk for a later time.

js/ad_manager_vast.js
```diff
@@ -483,6 +546,85 @@ OO.Ads.manager(function(_, $) {
...
+      };
+
+      this.initAd(this._properties['adWidth'], this._properties['adHeight'], viewMode,
+                  this._properties['adDesiredBitrate'], environmentVariables, creativeData);
+    }, this);
+
+    /**
+     * Initializes the ad by sending the data to the ad unit.
+     * @public
+     * @method VPaid#initAd
```
aeng7: The string before the # should be the class name. In this case, it should be Vast and not VPaid. If this function is used for VPAID only, please make note of it in the jsdoc and function name if necessary.

js/ad_manager_vast.js
```diff
@@ -483,6 +546,85 @@ OO.Ads.manager(function(_, $) {
...
+     * @param {int} desiredBitrate The bitrate for the ad
+     * @param {object} creativeData Contains the aditional ad parameters for the ad
+     * @param {object} environmentVars Contains the slot and videoSlot elements
+     */
+    this.initAd = function(width, height, viewMode, desiredBitrate, creativeData, environmentVars) {
+      if (!_isValidVPaid()) {
+        OO.log('VPaid Ad Unit is not valid.')
+        return;
+      }
+      currentAd.vpaidAd.initAd(width, height, viewMode, desiredBitrate, environmentVars, creativeData);
```
aeng7: Nice! vpaidAd is so much more readable than ad.ad :+1:

js/ad_manager_vast.js
nevermind89x: Right forgot to change all that when merging the code. Will do.

nevermind89x: Done.

js/ad_manager_vast.js
nevermind89x: Done

js/ad_manager_vast.js
```diff
@@ -618,16 +769,20 @@ OO.Ads.manager(function(_, $) {
...
-      if (currentAd && currentAd.ad && currentAd.ad.nextAdInPod) {
+      if (currentAd && currentAd.ad && currentAd.ad.nextAdInPod && !fromPause) {
         var metadata = currentAd.ad.nextAdInPod;
         if (metadata) {
           var ad = generateAd(metadata);
 
-          if (metadata.streamUrl != null || (ad.adType == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(metadata.streams))) {
+          if (metadata.streamUrl != null ||
+              (ad.adType == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(metadata.streams)) ||
+              currentAd.ad.data.adType === 'vpaid') {
```
aeng7: Can we introduce a utility function to check if an ad is a vpaid or not? I see this check used a few times.

js/ad_manager_vast.js
```diff
@@ -673,6 +829,9 @@ OO.Ads.manager(function(_, $) {
      */
     this.playAd = function(adWrapper) {
       currentAd = adWrapper;
+
+      var isVPaid = currentAd.ad.data.adType === 'vpaid';
```
aeng7: See note on making a utility function for this check.

js/ad_manager_vast.js
```diff
@@ -697,17 +851,27 @@ OO.Ads.manager(function(_, $) {
...
-      else {
+      } else {
         var streamUrl;
         if (adWrapper.ad && adWrapper.ad.streamUrl) {
           streamUrl = adWrapper.ad.streamUrl;
         }
         else if (adWrapper.streamUrl) {
           streamUrl = adWrapper.streamUrl;
+        } else {
+          streamUrl = '';
```
aeng7: What happens when streamUrl is the empty string? Will AMC get notified of an error?

js/ad_manager_vast.js
```diff
@@ -719,8 +883,15 @@ OO.Ads.manager(function(_, $) {
      * @param {object} adWrapper The current Ad's metadata
      */
     var initSkipAdOffset = _.bind(function(adWrapper) {
+      var isVPaid = adWrapper.ad.data.adType === 'vpaid';
+      var adSkippableState = false;
+      var skipOffset = '';
+      if (isVPaid) {
+        adSkippableState = adWrapper.vpaidAd.getAdSkippableState();
```
aeng7: Will getAdSkippableState always be defined and be a function?

js/ad_manager_vast.js
```diff
@@ -784,11 +972,17 @@ OO.Ads.manager(function(_, $) {
      * @param {boolean} failed If true, the ending of this ad was caused by a failure
      */
     var _endAd = _.bind(function(ad, failed) {
-      if (ad) {
-        if (ad.isLinear) {
+      if (currentAd && ad) {
+        var isLinear = ad.vpaidAd ? ad.vpaidAd.getAdLinear() : false;
+        if (ad.isLinear || isLinear) {
```
aeng7: Can we move the ad.isLinear check into the calculation for isLinear? This way we can have a check against only isLinear.

js/ad_manager_vast.js
```diff
@@ -784,11 +972,17 @@ OO.Ads.manager(function(_, $) {
...
           this.amc.notifyLinearAdEnded(ad.id);
+          if (transitionFromNonLinearVideo) {
+            this.amc.ui.transitionToMainContent(true, false);
+            transitionFromNonLinearVideo = false;
+            this.amc.notifyNonlinearAdEnded(ad.id);
+          }
```
aeng7: What does this block do?

js/ad_manager_vast.js
```diff
@@ -851,14 +1047,19 @@ OO.Ads.manager(function(_, $) {
...
     var addToTimeline = _.bind(function(metadata) {
+      if (!metadata) return;
       var timeline = [];
       var ad = generateAd(metadata);
+      var isVPaid = metadata.data && metadata.data.adType === 'vpaid';
 
-      if (metadata.streamUrl != null || (ad.adType == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(metadata.streams))) {
+      if (metadata.streamUrl != null ||
+          (ad.adType == this.amc.ADTYPE.LINEAR_VIDEO && !_.isEmpty(metadata.streams)) ||
+          (ad.adType === this.amc.ADTYPE.NONLINEAR_OVERLAY && !_.isEmpty(metadata.data.nonLinear.mediaFiles.url))) {
```
aeng7: This might need to get integrated with Doug's error handling changes. I recall errors for when streams or media files aren't defined. We need to check with Doug on this when we merge.

js/ad_manager_vast.js
```diff
@@ -909,7 +1110,7 @@ OO.Ads.manager(function(_, $) {
       var ooyalaClickUrl = amcAd.click_url;
       if (amcAd.isLinear) {
         adSpecificClickThroughUrl = amcAd.ad.data.linear.clickThrough;
-      } else if (amcAd.ad.data) {
+      } else {
```
aeng7: Why was this if removed? Was it causing an issue?

js/ad_manager_vast.js
```diff
@@ -926,7 +1127,10 @@ OO.Ads.manager(function(_, $) {
      * @param {object} amcAd The current ad data
      */
     this.pauseAd = function(amcAd) {
-      // No code required here as VTC will pause the ad
+      // Need to notify the ad unit that the player was paused
+      if (currentAd && currentAd.vpaidAd) {
+        currentAd.vpaidAd.pauseAd();
```
aeng7: Do we need to check if pauseAd is defined and a function?

js/ad_manager_vast.js
```diff
@@ -936,7 +1140,10 @@ OO.Ads.manager(function(_, $) {
      * @param {object} amcAd The current ad data
      */
     this.resumeAd = function(amcAd) {
-      // No code required here as VTC will resume the ad
+      // Need to notify the ad unit that the player was resumed
+      if (currentAd && currentAd.vpaidAd) {
+        currentAd.vpaidAd.resumeAd();
```
aeng7: Do we need to check if resumeAd is defined and a function?

aeng7:
Note: We should consider wrapping all calls to creatives (creative.skipAd, .pauseAd, .resumeAd, etc) to be wrapped within try catch statements. This way if the creative does something bad, our execution does not get disturbed.

js/ad_manager_vast.js
```diff
@@ -767,10 +948,17 @@ OO.Ads.manager(function(_, $) {
       if(params && params.code === this.amc.AD_CANCEL_CODE.TIMEOUT) {
         failedAd();
       } else {
+
+        if (params && params.code === this.amc.AD_CANCEL_CODE.SKIPPED && currentAd) {
+          if (currentAd.vpaidAd && typeof currentAd.vpaidAd.skipAd === 'function') {
+            // Notify Ad Unit that we are skipping the ad
+            currentAd.vpaidAd.skipAd();
```
aeng7: We could use a try/catch.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+        OO.log('VPaid: No valid media source, either is not a VPaid Ad or ad unit is not in javascript format.');
+        return;
+      }
+
+      if ($paramsNode.length) {
+        adParams = _cleanString($paramsNode.text());
+      }
+
+      var mediaFile = {
+        url: $mediaNode.text(),
```
aeng7: Should we use _cleanString on this?

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+     * Generates a parsed VPaid ad to load.
+     * This is only required for VPAID ads
+     * @private
+     * @method Vast#_getVPaidCreative
+     * @param {XMLDocument} adXml Current ad xml
+     * @param {string} Current vast version
+     * @param {object} adLoaded The ad loaded object and metadata
+     * @return {object} Parsed vpaid's metadata ad
+     */
+    var _getVpaidCreative = _.bind(function(adXml, version, adLoaded) {
```
aeng7: Can you add some comments in the function on what is happening? Will help with maintenance down the road. :+1:

js/ad_manager_vast.js
```diff
@@ -697,17 +851,27 @@ OO.Ads.manager(function(_, $) {
...
+          streamUrl = '';
         }
+
         this.amc.sendURLToLoadAndPlayNonLinearAd(adWrapper, adWrapper.id, streamUrl);
-        this.checkCompanionAds(adWrapper.ad);
+      }
+      this.checkCompanionAds(adWrapper.ad);
+      if (isVPaid) {
+        //Since IMA handles its own UI, we want the video player to hide its UI elements
+        this.amc.hidePlayerUi();
```
aeng7: There is an extra call to hidePlayerUI down in the _beginAd function. Can you check if one of these can be removed without affecting anything? Thanks!

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+     * @private
+     * @method Vast#_beginAd
+     */
+    var _beginAd = _.bind(function() {
+      var ad = currentAd.vpaidAd,
+          adLinear = ad.getAdLinear();
+
+      initSkipAdOffset(currentAd);
+      //Since IMA handles its own UI, we want the video player to hide its UI elements
+      this.amc.hidePlayerUi();
```
aeng7: There is an extra call to hidePlayerUI up in the playAd function. Can you check if one of these can be removed without affecting anything? Thanks!

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * Once Ad Playback stopped
+     * This is only required for VPAID ads
+     * @private
+     * @method Vast#_stopAd
+     */
+    var _stopAd = _.bind(function() {
+      if (currentAd && currentAd.vpaidAd) {
+        currentAd.vpaidAd.stopAd();
```
aeng7: We can use a try/catch statement here.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+     * @method Vast#_getTracking
+     * @param {object} parent DOM Element to look for tracking events
+     * @return {array} Array with tracking events and urls
+     */
+    this._getTracking = function(parent) {
+      var node, nodes, tracking, _i, _len;
+      tracking = [];
+      nodes = parent.getElementsByTagName('Tracking');
+      if (!nodes) {
+        return;
```
aeng7: Would returning an empty array here be better?

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+     * @param {string} type Event name to be send
+     */
+    this._sendTracking = function(type) {
+      var ad = prevAd ? prevAd : currentAd;
+      if (ad && ad.data) {
+        var tracking = ad.data.tracking,
+            currentEvent;
+        if (tracking) {
+          currentEvent = _.find(tracking, function(item, index) {
+            return item.event == type
```
aeng7: Missing semicolon

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * Listen and executes events sent by the ad unit
+     * This is only required for VPAID ads
+     * @private
+     * @method Vast#_onAdEvent
+     * @param {string} eventName Name of the event to process
+     */
+    var _onAdEvent = _.bind(function(eventName) {
+      switch(eventName) {
```
aeng7: We should consider try/catches to wrap any vpaidAd API calls. This will ensure that things like tracking don't get lost if the creative has an issue.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+      return element;
+    }, this);
+
+   /**
+     * Used to generate a frame to load ad media files.
+     * This is only required for VPAID ads
+     * @private
+     */
+    var _getFrame = function() {
+      iframe = document.createElement('iframe');
```
aeng7: Do iframes created by this function get disposed of properly after the ad is finished?

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+     * This is only required for VPAID ads
+     * @private
+     * @return {object} A DOM element with unique id.
+     */
+    var _createUniqueElement = _.bind(function() {
+      var element = null,
+          parent = this.amc.ui.playerSkinPluginsElement ?
+                              this.amc.ui.playerSkinPluginsElement[0] : this.amc.ui.pluginsElement[0];
+
+      element = document.createElement('div');
```
aeng7: Does this element get disposed of properly when the ad is finished?

js/ad_manager_vast.js
```diff
@@ -64,6 +64,69 @@ OO.Ads.manager(function(_, $) {
...
+                                            AD_VIDEO_FIRST_QUARTILE   : 'AdVideoFirstQuartile',
+                                            AD_VIDEO_MIDPOINT         : 'AdVideoMidpoint',
+                                            AD_VIDEO_THIRD_QUARTILE   : 'AdVideoThirdQuartile',
+                                            AD_USER_ACCEPT_INVITATION : 'AdUserAcceptInvitation',
+                                            AD_VOLUME_CHANGE          : 'AdVolumeChange',
+                                            AD_USER_MINIMIZE          : 'AdUserMinimize'
+                                          };
+
+    // If `false` prerolls won't load until intialPlay
+    this.preload                          = true;
```
pilievOoyala: In general we want to not preload ads. This also adds another hook into VAST to get ads to preload. So ideally we want to consolidate this into 1.

js/ad_manager_vast.js
```diff
@@ -483,6 +546,87 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * Initializes the ad by sending the data to the ad unit.
+     * @public
+     * This is only required for VPAID ads
+     * @method Vast#initAd
+     * @param {int} width Width of the slot where the ad will be placed
+     * @param {int} height Height of the slot where the ad will be placed
+     * @param {string} viewMode Can be either `normal` or `fullscreen`
+     * @param {int} desiredBitrate The bitrate for the ad
```
pilievOoyala: do we use int or number? You can fix this later if it's supposed to be 'number'.

js/ad_manager_vast.js
```diff
@@ -483,6 +546,87 @@ OO.Ads.manager(function(_, $) {
...
+     * @param {object} creativeData Contains the aditional ad parameters for the ad
+     * @param {object} environmentVars Contains the slot and videoSlot elements
+     */
+    this.initAd = function(width, height, viewMode, desiredBitrate, creativeData, environmentVars) {
```
pilievOoyala: Can you change this to initVpaidAd?  Please fix this everywhere else so it's clear, for code and documentation.

js/ad_manager_vast.js
```diff
@@ -697,17 +851,27 @@ OO.Ads.manager(function(_, $) {
...
+        } else {
+          streamUrl = '';
         }
+
         this.amc.sendURLToLoadAndPlayNonLinearAd(adWrapper, adWrapper.id, streamUrl);
-        this.checkCompanionAds(adWrapper.ad);
+      }
+      this.checkCompanionAds(adWrapper.ad);
+      if (isVPaid) {
+        //Since IMA handles its own UI, we want the video player to hide its UI elements
```
pilievOoyala: Is this comment copy/pasted from the IMA code? I don't see a reason to reference IMA here.

js/ad_manager_vast.js
```diff
@@ -1496,6 +1727,19 @@ OO.Ads.manager(function(_, $) {
     };
 
     /**
+     * Check wether or not a vpaid ad is valid by checking the ad type and make sure is VPaid
+     * This is only required for VPAID ads
+     * @method Vast#_isValidVPaid
+     * @private
+     * @return {boolean} VPaid validated value
+     */
+    var _isValidCreative = function(node, isLinear) {
```
pilievOoyala: please change function name to _isValidVPAID, not just in the comments.

js/ad_manager_vast.js
pilievOoyala: For statements like this you could create a "safeFunctionCall" method. AMC has code like that and so does the AF,if you need an example.

js/ad_manager_vast.js
pilievOoyala: Empty array is better but if you can't change it, then specify that it returns undefined in the @return documentation.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+      return tracking;
+    };
+
+    /**
+     * Send error.
+     * This is only required for VPAID ads
+     * @public
+     * @method Vast#_sendError
+     */
+    this._sendError = function() {
```
pilievOoyala: If it's VPAID only, make it _sendVPAIDError

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+      }
+    };
+
+    /**
+     * Send impressions.
+     * This is only required for VPAID ads
+     * @public
+     * @method Vast#_sendImpressions
+     */
+    this._sendImpressions = function() {
```
pilievOoyala: Same thing here, make it _sendVPAIDImpressions. Please repeat that for these rest of the VPAID only functions.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+    };
+
+    /**
+     * Send tracking events.
+     * This is only required for VPAID ads
+     * @public
+     * @method Vast#_sendTracking
+     * @param {string} type Event name to be send
+     */
+    this._sendTracking = function(type) {
```
pilievOoyala: this is marked as @public but the function name starts with '_'. Do you need to change it to @private? Also, if it's private why is it part of 'this'?  Same applies to other functions in this section.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+
+        case VPAID_EVENTS.AD_ERROR:
+          OO.log('VPaid: Ad unit error: ' + arguments[1]);
+          this._sendTracking('error');
+          this._sendError();
+          failedAd();
+        break;
+
+        case VPAID_EVENTS.AD_DURATION_CHANGE:
+          var remainingTime = currentAd.vpaidAd.getAdRemainingTime();
```
aeng7: remainingTime is not used. If we do not need this, please remove.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2127,625 @@ OO.Ads.manager(function(_, $) {
...
+
+    /**
+     * Starts the click-to-linear ad
+     * This is only required for VPAID ads
+     * @private
+     * @method Vast#_beginAd
+     */
+    var _beginAd = _.bind(function() {
+      var ad = currentAd.vpaidAd,
+          adLinear = ad.getAdLinear();
```
aeng7: Is the adLinear variable used for anything?

js/ad_manager_vast.js
```diff
@@ -564,15 +564,25 @@ OO.Ads.manager(function(_, $) {
...
+      } catch (e) {
+        OO.log("VPAID 2.0: error while getting vpaid creative - " + e)
+      }
+
 
       // Subscribe to ad unit events
       for (eventName in VPAID_EVENTS) {
-        currentAd.vpaidAd.subscribe(_.bind(_onAdEvent, this, VPAID_EVENTS[eventName]),
+        try{
+          currentAd.vpaidAd.subscribe(_.bind(_onVpaidAdEvent, this, VPAID_EVENTS[eventName]),
```
pilievOoyala: can you use the _safeFunctionCall here as well?

js/ad_manager_vast.js
```diff
@@ -2488,70 +2499,70 @@ OO.Ads.manager(function(_, $) {
...
+          this.sendVpaidError();
           failedAd();
         break;
 
         case VPAID_EVENTS.AD_DURATION_CHANGE:
-          var remainingTime = currentAd.vpaidAd.getAdRemainingTime();
-          if (!duration) {
-            _stopAd();
+          var remainingTime = _safeFunctionCall(currentAd.vpaidAd, "getAdRemainingTime");
+          if (remainingTime <= 0) {
```
pilievOoyala: nice catch!

js/ad_manager_vast.js
```diff
@@ -2604,16 +2615,21 @@ OO.Ads.manager(function(_, $) {
      * @return {boolean} VPaid validated value
      */
     var _isValidVPaid = _.bind(function() {
-      var creative = currentAd.vpaidAd,
-          vpaidVersion = parseInt(creative.handshakeVersion('2.0')),
-          isValid = true;
+      var vpaidVersion = null;
+      try{
+        vpaidVersion = parseInt(currentAd.vpaidAd.handshakeVersion('2.0'));
```
pilievOoyala: can you use _safeFunctionCall here?

pilievOoyala: Also, do you want parseInt or parseFloat? since we could have minor updates to the standard later.

pilievOoyala: maybe rename the var to vpaidMajorVersion.

js/ad_manager_vast.js
```diff
@@ -1875,6 +2142,664 @@ OO.Ads.manager(function(_, $) {
...
+      var isValid = true;
+
+      if (vpaidVersion !== 2) {
+        OO.log('VPaid Ad Unit version is not supported.');
+        isValid = false;
+      }
+
+      var requiredFunctions = ['handshakeVersion', 'initVpaidAd', 'startAd', 'stopAd', 'skipAd', 'resizeAd',
+                               'pauseAd', 'resumeAd', 'expandAd', 'collapseAd', 'subscribe', 'unsubscribe'];
+      _.each(requiredFunctions, function(fn) {
```
nevermind89x: oh,  My bad here we should check against ad unit functions. also it should be initAd in the array.

PR #70, opened 2016-03-22
[PBI-1597]

aeng7:
-Workaround of a size change notification issue so that IMA ads and VPAID 2.0 standalone ads resize properly


pilievOoyala:
LGTM.

PR #71, opened 2016-03-22
Vast vpaid

aeng7:
-fixed end of ad flow for VPAID 2.0 ads. We will wait for AD_STOPPED from creative before proceeding
-removed extra iframes when VPAID 2.0 ad finishes
-removed video restrictions for VAST ads
-renamed iframe to vpaidIframe for clarification


pilievOoyala:
LGTM.

PR #73, opened 2016-03-23
[PBI-1599]

aeng7:
-added a workaround to handle an issue that arises when the user clicks on “Done” on iPhone devices, leaving the native video player.


pilievOoyala:
LGTM

PR #75, opened 2016-03-24
[PBI-1608]

aeng7:
-fixed an issue where variables were out of order when calling initVpaidAd. We now properly follow the VPAID 2.0 spec
-fixed an issue where exceptions in creatives were not properly encapsulated by a try/catch in safeFunctionCall
-note that Francisco is currently working on unit tests but will not make the March build. The unit tests will test for these issues

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -608,7 +608,7 @@ OO.Ads.manager(function(_, $) {
       };
 
       this.initVpaidAd(this._properties['adWidth'], this._properties['adHeight'], viewMode,
-                  this._properties['adDesiredBitrate'], environmentVariables, creativeData);
+                  this._properties['adDesiredBitrate'], creativeData, environmentVariables);
     }, this);
 
     /**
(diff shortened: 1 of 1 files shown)


js/ad_manager_vast.js
```diff
@@ -2831,7 +2831,7 @@ OO.Ads.manager(function(_, $) {
         }
       } catch (err) {
         OO.log("VPAID 2.0: ",
-          "function '" + func + "' threw exception -",
+          "function '" + funcName + "' threw exception -",
```
angieisrani: this looks good

js/ad_manager_vast.js
```diff
@@ -608,7 +608,7 @@ OO.Ads.manager(function(_, $) {
       };
 
       this.initVpaidAd(this._properties['adWidth'], this._properties['adHeight'], viewMode,
-                  this._properties['adDesiredBitrate'], environmentVariables, creativeData);
+                  this._properties['adDesiredBitrate'], creativeData, environmentVariables);
```
angieisrani: I think I exlplained the issue wrong, is this working for you?

aeng7: yep I tested and its working. I modified the code to match the spec: http://www.iab.com/wp-content/uploads/2015/06/VPAID_2_0_Final_04-10-2012.pdf

initAd(width : Number, height : Number, viewMode : String, desiredBitrate : Number,
[creativeData : String], [environmentVars : String]) : void

PR #77, opened 2016-03-30
[PBI-1636] Add unit tests for vpaid

nevermind89x:
Please review :shipit:  I was adding some repeated tests for vast, until I realized it wasn't required, so only added vpaid specific stuff to the tests.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -553,25 +553,25 @@ OO.Ads.manager(function(_, $) {
     /**
      * Callback when the media file is loaded. Once is loaded we can initialize the ad
      * This is only required for VPAID ads
-     * @private
+     * @public
      */
-    var _initializeAd = _.bind(function() {
+    this.initializeAd = _.bind(function() {
       var eventName,
           environmentVariables,
           viewMode,
           creativeData = {};
 
       currentAd.data = currentAd.ad.data;
-      if (!this.amc.ui.adVideoElement[0]) {
+      if (!this.amc.ui.adVideoElement) {
         this.amc.ui.createAdVideoElement({'mp4' : ''}, vpaidVideoRestrictions);
       }
-      if (typeof vpaidIframe.contentWindow.getVPAIDAd !== 'function') {
+      if (typeof vpaidIframe.contentWindow.getVPAIDAd !== 'function' && !this.testMode) {
         OO.log('VPAID 2.0: Required function getVPAIDAd() is not defined.');
         return;
       }
 
       try{
-        currentAd.vpaidAd = vpaidIframe.contentWindow.getVPAIDAd();
+        currentAd.vpaidAd = this.testMode ? global.vpaid.getVPAIDAd() : vpaidIframe.contentWindow.getVPAIDAd();
       } catch (e) {
         OO.log("VPAID 2.0: error while getting vpaid creative - " + e)
       }
@@ -1772,7 +1772,7 @@ OO.Ads.manager(function(_, $) {
      * @return {boolean} VPaid validated value
      */
     var _isValidVpaidCreative = function(node, isLinear) {
-      var apiFramework = node.attr('apiFramework') === 'VPAID';
+      var apiFramework = (node.attr('apiFramework') || node.attr('apiframework')) === 'VPAID';
       var creativeType = isLinear ? node.attr('type') : node.find('StaticResource').attr('creativeType');
       return apiFramework && creativeType === 'application/javascript';
     };
@@ -2311,7 +2311,7 @@ OO.Ads.manager(function(_, $) {
       this.amc.notifyLinearAdStarted(currentAd.id, {
         name: currentAd.data.title,
         duration : _safeFunctionCall(ad, "getAdDuration"),
-        clickUrl: _hasClickUrl(currentAd),
+        clickUrl: currentAd.ad.data.nonLinear.nonLinearClickThrough.length > 0,
         indexInPod: currentAd.ad.sequence,
         skippable : _safeFunctionCall(ad, "getAdSkippableState")
       });
@@ -2670,10 +2670,10 @@ OO.Ads.manager(function(_, $) {
         isValid = false;
       }
 
-      var requiredFunctions = ['handshakeVersion', 'initVpaidAd', 'startAd', 'stopAd', 'skipAd', 'resizeAd',
+      var requiredFunctions = ['handshakeVersion', 'initAd', 'startAd', 'stopAd', 'skipAd', 'resizeAd',
                                'pauseAd', 'resumeAd', 'expandAd', 'collapseAd', 'subscribe', 'unsubscribe'];
       _.each(requiredFunctions, function(fn) {
-        if (!fn && typeof fn !== 'function') {
+        if (typeof currentAd.vpaidAd[fn] !== 'function') {
           isValid = false;
           OO.log('VPaid Ad Unit is missing function: ' + fn);
         }
@@ -2722,7 +2722,7 @@ OO.Ads.manager(function(_, $) {
     var _onIframeLoaded = _.bind(function() {
       var loader = vpaidIframe.contentWindow.document.createElement('script');
       loader.src = _cleanString(currentAd.ad.mediaFile.url);
-      loader.onload = _initializeAd;
+      loader.onload = this.initializeAd;
       loader.onerror = this.destroy;
       vpaidIframe.contentWindow.document.body.appendChild(loader);
     }, this);
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -2299,22 +2299,25 @@ OO.Ads.manager(function(_, $) {
      * @method Vast#_beginVpaidAd
      */
     var _beginVpaidAd = _.bind(function() {
-      var ad = currentAd.vpaidAd;
-      //TODO: Is this used for anything?
-      var adLinear = _safeFunctionCall(ad, "getAdLinear");
+      if (currentAd && currentAd.ad && currentAd.ad.data) {
+        var ad = currentAd.vpaidAd;
+        var clickthru = currentAd.ad.data.nonLinear ? currentAd.ad.data.nonLinear.nonLinearClickThrough : '';
+        //TODO: Is this used for anything?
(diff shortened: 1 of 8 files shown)


js/ad_manager_vast.js
```diff
@@ -2311,7 +2311,7 @@ OO.Ads.manager(function(_, $) {
       this.amc.notifyLinearAdStarted(currentAd.id, {
         name: currentAd.data.title,
         duration : _safeFunctionCall(ad, "getAdDuration"),
-        clickUrl: _hasClickUrl(currentAd),
+        clickUrl: currentAd.ad.data.nonLinear.nonLinearClickThrough.length > 0,
```
pilievOoyala: can we ad in a check for currentAd being truthy and currentAd.ad?

pilievOoyala: Also, do some null checks for this big clickUrl value.

js/ad_manager_vast.js
```diff
@@ -2479,6 +2479,8 @@ OO.Ads.manager(function(_, $) {
       switch(eventName) {
         case VPAID_EVENTS.AD_LOADED:
           adLoaded = true;
+          //For VPAID we need to check for companions after ad is created
+          this.checkCompanionAds(currentAd.ad);
```
pilievOoyala: do a truthy check for currentAd here as well.

test/unit-tests/vast_test.js
```diff
@@ -1721,4 +1761,281 @@ describe('ad_manager_vast', function() {
     vastAdManager.loadMetadata({"tagUrl": "http://blahblah"}, {}, content);
     expect(vastAdManager.vastUrl).to.be("http://blahblah");
   });
+
+  it('VPAID 2.0: Should parse VPAID linear creative', function() {
+    vpaidInitialize();
+    var ad = amc.timeline[0];
+    expect(ad).to.be.an('object');
+    expect(ad.duration).to.eql(16);
```
pilievOoyala: since we have number values coming through. Are you checking if string values work as well?  So <myTag duration=15>  and <myTag duration="15">

pilievOoyala:
Looks good, can you add some negative testing, where you are missing info in the xml or the value is bad?

test/unit-tests/vast_test.js
```diff
@@ -1721,4 +1761,281 @@ describe('ad_manager_vast', function() {
...
+    };
+    vpaidInitialize();
+    var ad = amc.timeline[0];
+    vastAdManager.playAd(ad);
+    vastAdManager.initializeAd();
+    expect(hidePlayerUi).to.be(true);
+  });
+
+  it('VPAID 2.0: Should resize ad unit on size changed', function() {
+    var hidePlayerUi = false;
```
aeng7: This var is unused.

test/unit-tests/vast_test.js
```diff
@@ -1721,4 +1761,281 @@ describe('ad_manager_vast', function() {
...
+    expect(ad.vpaidAd.properties.height).to.be(100);
+    vastAdManager._slot.offsetWidth = 200;
+    vastAdManager._slot.offsetHeight = 200;
+    amc.publishPlayerEvent(amc.EVENTS.SIZE_CHANGED);
+    expect(ad.vpaidAd.properties.width).to.be(200);
+    expect(ad.vpaidAd.properties.height).to.be(200);
+  });
+
+  it('VPAID 2.0: Should resize ad unit on fullscreen change', function() {
+    var hidePlayerUi = false;
```
aeng7: This var is unused.

js/ad_manager_vast.js
```diff
@@ -2670,10 +2672,10 @@ OO.Ads.manager(function(_, $) {
         isValid = false;
       }
 
-      var requiredFunctions = ['handshakeVersion', 'initVpaidAd', 'startAd', 'stopAd', 'skipAd', 'resizeAd',
+      var requiredFunctions = ['handshakeVersion', 'initAd', 'startAd', 'stopAd', 'skipAd', 'resizeAd',
```
aeng7: nice catch!

js/ad_manager_vast.js
```diff
@@ -2670,10 +2672,10 @@ OO.Ads.manager(function(_, $) {
...
                                'pauseAd', 'resumeAd', 'expandAd', 'collapseAd', 'subscribe', 'unsubscribe'];
       _.each(requiredFunctions, function(fn) {
-        if (!fn && typeof fn !== 'function') {
+        if (typeof currentAd.vpaidAd[fn] !== 'function') {
```
aeng7: Can we check for currentAd to be truthy?

test/unit-tests/vast_test.js
```diff
@@ -1721,4 +1761,281 @@ describe('ad_manager_vast', function() {
...
+    var ad = amc.timeline[0];
+    vastAdManager.playAd(ad);
+    vastAdManager.initializeAd();
+    expect(ad.vpaidAd.properties.width).to.be(100);
+    expect(ad.vpaidAd.properties.height).to.be(100);
+    vastAdManager._slot.offsetWidth = 200;
+    vastAdManager._slot.offsetHeight = 200;
+    amc.publishPlayerEvent(amc.EVENTS.SIZE_CHANGED);
+    expect(ad.vpaidAd.properties.width).to.be(200);
+    expect(ad.vpaidAd.properties.height).to.be(200);
```
aeng7: Can we use different values for width and height? This will allow us to detect if there is a width/height mixup in our logic in the future.

test/unit-tests/vast_test.js
```diff
@@ -1721,4 +1761,281 @@ describe('ad_manager_vast', function() {
...
+    var ad = amc.timeline[0];
+    vastAdManager.playAd(ad);
+    vastAdManager.initializeAd();
+    expect(ad.vpaidAd.properties.width).to.be(100);
+    expect(ad.vpaidAd.properties.height).to.be(100);
+    vastAdManager._slot.offsetWidth = 200;
+    vastAdManager._slot.offsetHeight = 200;
+    amc.publishPlayerEvent(amc.EVENTS.FULLSCREEN_CHANGED);
+    expect(ad.vpaidAd.properties.width).to.be(200);
+    expect(ad.vpaidAd.properties.height).to.be(200);
```
aeng7: Can we use different values for width and height? This will allow us to detect if there is a width/height mixup in our logic in the future.

test/unit-tests/vast_test.js
nevermind89x: Alex didn't get what you mean by use different values? You mean to have a test for the case when values don't match? (something like `no.to.be(300)`)?

aeng7: Yep, like have a width of 300 and a height of 200.

nevermind89x: got it.

test/unit-tests/vast_test.js
nevermind89x: Not really sure we should check for strings, the duration tag is like this: `<Duration>00:00:16</Duration>` 

Added a validation for the case: `<Duration>16</Duration>`

nevermind89x:
All the changes are done. Added few more things.

@pilievOoyala, for the negative testing we have no validations (for vast/vpaid) for missing values. regarding on checking no `duration` not sure what's the expected behavior should be if duration is missing?? Right now for both it is being saved as NaN (if empty duration) but ads are playing fine.

js/ad_manager_vast.js
```diff
@@ -561,7 +561,14 @@ OO.Ads.manager(function(_, $) {
           viewMode,
           creativeData = {};
 
+    if (_isVpaidAd(currentAd)) {
       currentAd.data = currentAd.ad.data;
+
+      var mediaFileUrl = _cleanString(currentAd.ad.mediaFile.url);
```
pilievOoyala: Can you check (currentAd && currentAd.ad && currentAd.ad.mediaFile)?  Otherwise this could potential crash. Or you can do a try/catch

js/ad_manager_vast.js
```diff
@@ -2299,22 +2308,25 @@ OO.Ads.manager(function(_, $) {
...
-      this.amc.notifyLinearAdStarted(currentAd.id, {
-        name: currentAd.data.title,
-        duration : _safeFunctionCall(ad, "getAdDuration"),
-        clickUrl: currentAd.ad.data.nonLinear.nonLinearClickThrough.length > 0,
-        indexInPod: currentAd.ad.sequence,
-        skippable : _safeFunctionCall(ad, "getAdSkippableState")
-      });
+      if (_isVpaidAd(currentAd)) {
+        var ad = currentAd.vpaidAd;
+        var clickthru = currentAd.ad.data.nonLinear ? currentAd.ad.data.nonLinear.nonLinearClickThrough : '';
```
pilievOoyala: can you do the truthy checks for all these variables.
(currentAd && currentAd.ad && currentAd.ad.data)

js/ad_manager_vast.js
```diff
@@ -2299,22 +2308,25 @@ OO.Ads.manager(function(_, $) {
...
+        //TODO: Is this used for anything?
+        var adLinear = _safeFunctionCall(ad, "getAdLinear");
+
+        initSkipAdOffset(currentAd);
+        //Since a VPAID 2.0 ad handles its own UI, we want the video player to hide its UI elements
+        this.amc.hidePlayerUi();
+        this.amc.notifyPodStarted(currentAd.id, currentAd.ad.adPodLength);
```
pilievOoyala: do truthy check here too.
(currentAd &&currentAd.ad)

js/ad_manager_vast.js
nevermind89x: Sure well _isVPaidAd already checks for ad && ad.ad && ad.ad.data, but will make sure to check for mediaFile, which is mostly only for testing purposes, because basically we would never get here if mediaFile is empty/null, since this function is only loaded when `script.onload=initialize` otherwise if empty we would go to `script.onerror`

js/ad_manager_vast.js
nevermind89x: _isVPaidAd already checks for ad && ad.ad && ad.ad.data

js/ad_manager_vast.js
nevermind89x: Same here, already checked

pilievOoyala:
LGTM

PR #84, opened 2016-04-11
[PBI-1529] Enable VAST to catch MIME types for HLS

dchauOoyala:


aeng7:
Can we add a unit test? LGTM otherwise

pilievOoyala:
LGTM

PR #85, opened 2016-04-11
[PBI-1666]

aeng7:
-fixed an issue where we do not provide the correct shared video element on iOS to the IMA SDK when the original video element is destroyed and recreated


angieisrani:
LGTM

PR #86, opened 2016-04-13
[PBI-1608]

aeng7:
-VPAID in the Vast ad manager will now provide ad level video restrictions in the ad object that is provided to the AMC


aeng7:
Must be merged after mjolnir PR 22

angieisrani:
LGTM

PR #88, opened 2016-04-15
Pbi 1648

pilievOoyala:
Taking out preloading of vast ads. Ads now load only when they are about to play, using a placeholder linear-overlay in the timeline(similar to what we do in freewheel).  And unit tests were updated to reflect the changes.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -891,7 +891,6 @@ OO.Ads.manager(function(_, $) {
      * @param  {object} adWrapper An object of type AdManagerController.Ad containing ad metadata
      */
     var _playLoadedAd = _.bind(function(adWrapper) {
-      this.amc.ui.createAdVideoElement(adWrapper.streams);
       var isVPaid = _isVpaidAd(currentAd);
 
       // When the ad is done, trigger callback
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -79,7 +79,7 @@ OO.Ads.manager(function(_, $) {
     // ad settings
     var transitionFromNonLinearVideo    = false;
     var adLoaded                        = false;
-    var adRequestType                   = 'adRequest';
+    var AD_REQUEST_TYPE                 = 'adRequest';
 
     // VPAID variables
     this._slot                          = null;
@@ -746,9 +746,8 @@ OO.Ads.manager(function(_, $) {
         override = true;
         ad.tag_url = this.adURLOverride;
       }
-      var time = typeof ad.time !== 'undefined' ? ad.time : ad.position;
-      this.currentAdBeingLoaded = ad;
 
+      this.currentAdBeingLoaded = ad;
       var url = typeof ad.url !== 'undefined' && !override ? ad.url : ad.tag_url;
       this.loadUrl(url);
       loadedAds = true;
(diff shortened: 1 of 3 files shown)


js/ad_manager_vast.js
```diff
@@ -739,34 +731,27 @@ OO.Ads.manager(function(_, $) {
...
-            loadedAds = true;
-          }
-        }
+      var ad = amcAd.ad;
+      this.amc.notifyPodStarted(amcAd.id, 1);
+      if (this.adURLOverride) {
+        override = true;
+        ad.tag_url = this.adURLOverride;
+      }
+      var time = typeof ad.time !== 'undefined' ? ad.time : ad.position;
```
aeng7: Is var time used for anything?

js/ad_manager_vast.js
```diff
@@ -867,8 +877,18 @@ OO.Ads.manager(function(_, $) {
...
-      currentAd = adWrapper;
+      if (adWrapper) {
+        currentAd = adWrapper;
+        if (currentAd.isAdRequest) {
+          loadAd(currentAd);
+        } else {
+          playLoadedAd(adWrapper);
+        }
+      }
+    }
```
aeng7: Missing semicolon

js/ad_manager_vast.js
pilievOoyala: not anymore, good catch.

js/ad_manager_vast.js
```diff
@@ -867,8 +877,18 @@ OO.Ads.manager(function(_, $) {
...
 
+    var playLoadedAd = _.bind(function(adWrapper) {
```
aeng7: The vast ad manager uses underscore notation for private members. Can we change this to _playLoadedAd. Also, can we add a jsdoc for this?

js/ad_manager_vast.js
```diff
@@ -801,7 +786,32 @@ OO.Ads.manager(function(_, $) {
...
+          };
+
+          if (ad.position_type == 't') {
+            adData.position = ad.time/1000;
+          } else if (ad.position_type == 'p') {
+            //TODO
+            //adData.position =
+          }
+          var amcAd = new this.amc.Ad(adData);
+          amcAd.isAdRequest = true;
```
aeng7: The IMA ad manager uses an ad request type "adRequest" to mark an ad as an ad request. We can do the same thing here to maintain consistency across ad managers.

js/ad_manager_vast.js
```diff
@@ -1921,6 +1941,13 @@ OO.Ads.manager(function(_, $) {
       } else {
         failedAd();
       }
+
+      if (currentAd && currentAd.isAdRequest)
```
aeng7: Is an ad request ad able to make it to the handleAds function? handleAds is only called after we receive the Vast response in the onVastResponse function.

test/unit-test-helpers/mock_amc.js
```diff
@@ -42,7 +42,8 @@ fake_amc = function() {
     playerSkinPluginsElement : $("<div>", {}),
     adWrapper : $("<div>", {}),
     transitionToAd: function(){},
-    transitionToMainContent: function(){}
+    transitionToMainContent: function(){},
+    createAdVideoElement: function() {}
```
aeng7: There was a recent change that removed the calling of createAdVideoElement by the vast ad manager. Angie and I had a discussion and agreed this should not be an API for the ad plugins to use. Can you try removing this? If this breaks some unit tests, we can merge this branch with master prior to pushing.

test/unit-tests/vast_test.js
```diff
@@ -120,13 +119,14 @@ describe('ad_manager_vast', function() {
...
-    initalPlay();
+    vastAdManager.loadMetadata(server, {}, content);
+    initialPlay();
+    vastAdManager.initialPlay();
     xml = xml || vpaidLinearXML;
     vastAdManager.onVastResponse(preroll, xml);
   };
 
-  var initalPlay = function() {
+  var initialPlay = function() {
```
aeng7: Nice catch! :D

test/unit-tests/vast_test.js
```diff
@@ -443,28 +383,31 @@ describe('ad_manager_vast', function() {
...
+      "html5_ad_server": "http://blah"}, {}, content);
+    expect(amc.timeline.length).to.be(1);
+
+    initialPlay();
+    vastAdManager.initialPlay();
     vastAdManager.onVastResponse(vast_ad,'<VAST></VAST>');
+
     expect(errorType.length > 0).to.be(true);
-    expect(amc.timeline.length).to.be(0);
+    expect(amc.timeline.length).to.be(1);
```
aeng7: Do we check the 2nd element in the array for the unit tests because the first element is the ad request ad?

pilievOoyala: Yes. I'd like to revamp these tests but we should do the moving of the AMC into this repo first.  There is too much emulation going on.

aeng7: Sounds good.

aeng7:
LGTM. Waiting on the mjolnir side.

PR #92, opened 2016-04-26
[PBW-5002] In case a timing conflict with VTC occurs with VPAID ad lo…

angieisrani:
…ad, force the ad to load and play

The problem is that the VPAID ad manager is manually doing stuff to the video element before it's ready for usage.  To fix that, force play if required.

Also fix a couple style issues


chibzu:
Looks good to me, @aeng7 for reference.

aeng7:
LGTM

PR #94, opened 2016-04-29
Pbi 1734

pilievOoyala:
Fixed error handling code during vast parsing errors. The change to only load ads when needed, exposed this.  Also fixed up several other places where we need to do null checks because the unit test uncovered it.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -803,7 +803,11 @@ OO.Ads.manager(function(_, $) {
           } else {
             var adPod = adPodPrimary;
             adPodPrimary = null;
-            this.amc.notifyPodEnded(adPod.id);
+            if(adPod) {
+              this.amc.notifyPodEnded(adPod.id);
+            } else if (badAd) {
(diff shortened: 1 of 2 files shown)


js/ad_manager_vast.js
```diff
@@ -789,23 +789,27 @@ OO.Ads.manager(function(_, $) {
...
+        //otherwise end the ad pod
+        if (metadata) {
+          var ad = generateAd(metadata);
+          this.amc.forceAdToPlay(this.name, ad.ad, ad.adType, ad.streams);
+        } else {
+          var adPod = adPodPrimary;
+          adPodPrimary = null;
+          if(adPod) {
+             this.amc.notifyPodEnded(adPod.id);
+             console.log("derp1");
```
aeng7: These should be easter eggs from now on :D

pilievOoyala: Man I am terrible at finding my own debugging code.

aeng7:
LGTM

PR #96, opened 2016-05-04
Delete ad_manager_videoplaza.js

apesant:
As we now have an ad manager for Pulse, it would be good to remove the videoplaza-ads-manager file.


angieisrani:
LGTM

PR #101, opened 2016-05-16
Add Ooyala Pulse plugin v4.3.3

eshaglund:
Fix for PSDK-366: HTML5 Player on Android showing black screen when playing content.

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -1,7 +1,7 @@
 /*
  * Pulse ad player ad manager
  *
- * version 1.0
+ * version 4.3.3
  */
 
 (function(_, $)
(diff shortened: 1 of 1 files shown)


js/pulse.js
```diff
@@ -806,7 +806,8 @@
          */
         this.sharedElementGive = function() {
             setTimeout(function(){
-                _adManager.ui.ooyalaVideoElement[0].play();
+                _adManager.sharedVideoElement.style.display = "block";
+                _adManager.sharedVideoElement.play();
             }, 100);
             _adManager.sharedVideoElement.style.visibility ="hidden";
```
angieisrani: @pilievOoyala I think the plugin shouldn't need to set visibility since vtc is doing that - are you able to verify this?

pilievOoyala: Correct, we shouldn't need to edit the visibility.  Have you tried debugging in to the video controller and seeing if there is a bug there?  If this plugin edits the visibility and the VTC does, then we can get out of sync. Which is what I suspect may be the root cause.

pilievOoyala: Also, in the future please feel free to remind us to review these changes so they don't get forgotten on our side.

eshaglund:
Changes merged in another PR

PR #103, opened 2016-05-18
phi-1749 Adding in page level setting for google ima 'useGoogleAdUI'.…

pilievOoyala:
…  It defaults to false but if you set it to true.  We will use Google's UI while playing ads.  This was added because we disable our ad UI and we are using this ad ui instead now(if needed).

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -147,6 +147,7 @@ require("../html5-common/js/utils/utils.js");
         this.adPlaybackStarted = false;
         this.vcPlayRequested = false;
         this.savedVolume = -1;
+        this.useGoogleAdUI = false;
 
         //flag to track whether ad rules failed to load
         this.adRulesLoadError = false;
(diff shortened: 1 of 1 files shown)


js/google_ima.js
```diff
@@ -147,6 +147,7 @@ require("../html5-common/js/utils/utils.js");
         this.adPlaybackStarted = false;
         this.vcPlayRequested = false;
         this.savedVolume = -1;
+        this.useGoogleAdUI = false;
```
angieisrani: defaults to false?

angieisrani:
phi - PlayHack Integration

js/google_ima.js
pilievOoyala: Because it's unproven and only 1 customer is requesting this. It'll be off by default.

angieisrani: Sorry quick q though - don't we want to use Google's UI elements by default and only use ours if if specified?  or does this mean to use Ooyala UI for google ads?  I think the name confused me

pilievOoyala: this is purely a flag to turn on Google's UI for ads.  Our is turned off permanently for Google IMA ads.  To not block ui elements like the skip button.  The skip button is always there, but this flag enabled stuff like ad countdown and their 'learn more' button.

angieisrani: So by default we don't want to show the IMA-rendered learn more button nor our own?

pilievOoyala: For the moment, yes.  If you don't have the ui enabled, it will do a clickthrough if you click on the video.  We may change the default later but since this is a very last minute change, before the cut, we are playing it safe and keeping the behavior we have at the moment.

angieisrani:
Okay LGTM - let's document this param

PR #106, opened 2016-05-24
[pbi-1743]

aeng7:
-expose a page level parameter to allow people to request the INSECURE vpaid mode in the IMA ad manager


pilievOoyala:
LGTM. Just make sure Docs knows about this new param.

PR #111, opened 2016-06-14
pbi-1814 Putting in example of how to get ID3 tag information in an a…

pilievOoyala:
…d manager.  PLEASE merge this after the mjolnir checkin.

### DIFF
--- a/js/ad_manager_template.js
+++ b/js/ad_manager_template.js
@@ -220,8 +220,18 @@ OO.Ads.manager(function(_, $) {
 
     };
 
-    this.onVideoTagFound = function() {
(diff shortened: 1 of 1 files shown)


js/ad_manager_template.js
```diff
@@ -217,6 +220,10 @@ OO.Ads.manager(function(_, $) {
 
     };
 
+    this.onVideoTagFound = function() {
```
dchauOoyala: Should add jsdoc here. Also, would it be better to add the arguments to the function signature?

dchauOoyala:
LGTM

PR #116, opened 2016-06-23
[PBW-4530] After skipping a postroll, control bar on end screen show position of ad when it was skipped

mityaha:
Added adsPlayed placeholder public method


chibzu:
Looks good.

PR #117, opened 2016-06-23
[PBI-1824b] Fix error in caching logic

dchauOoyala:
- this.currentAd.id3AdId was undefined

### DIFF
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -170,6 +170,7 @@ OO.Ads.manager(function(_, $)
         this.currentAd = ad;
         if (ad.ad)
         {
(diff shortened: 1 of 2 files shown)


js/ssai_pulse.js
```diff
@@ -170,6 +170,7 @@ OO.Ads.manager(function(_, $)
         this.currentAd = ad;
         if (ad.ad)
         {
+          this.currentAd.ad.id3AdId = this.currentId3Object.adId;
```
pilievOoyala: need to change this to this.ad.ad.id3AdId or change the if statement .

pilievOoyala:
Looks good to me.

PR #122, opened 2016-07-08
[PBF-661] Adds encoding for Akamai live/remote for secure tokens.

brianfwong:


brianfwong:
@agunawan @bobooyala PBF-661 - updates pointer to html5-common, which has new constant for HD2, needed by akamai and analytics. Keeps submodules consistent across repos.

PR #123, opened 2016-07-08
[PBI-1847]

dchauOoyala:
- Add player configuration parameter: "cacheBuster". Default true. Only
an explicit value of "false" will turn off cacheBuster functionality
- Add function to replace CACHEBUSTING macro in tracking urls before pinging

### DIFF
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -144,9 +144,21 @@ OO.Ads.manager(function(_, $)
 
       if (adManagerMetadata)
       {
-        if (adManagerMetadata["cacheBuster"] === "false")
+        if (adManagerMetadata["cacheBuster"])
         {
-          bustTheCache = false;
+          if (adManagerMetadata["cacheBuster"] === "true")
+          {
+            bustTheCache = true;
+          }
+          else if (adManagerMetadata["cacheBuster"] === "false")
+          {
+            bustTheCache = false;
+          }
+          else
+          {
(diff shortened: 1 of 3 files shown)


js/ssai_pulse.js
```diff
@@ -899,6 +922,26 @@ OO.Ads.manager(function(_, $)
...
+     * @private
+     * @method SsaiPulse#_cacheBuster
+     * @param {string[]} urls The array of URLs
+     * @returns {string[]} The new array of URLs.
+     */
+    var _cacheBuster = _.bind(function(urls)
+    {
+      for (var i = 0; i < urls.length; i++)
+      {
+        var regex = /%5bcachebusting%5d|\[cachebusting\]/i;
```
aeng7: Can we try decodeURIComponent instead of %5b and %5d?

aeng7: Can we also store the [cachebusting] value in a var?

js/ssai_pulse.js
```diff
@@ -138,6 +141,14 @@ OO.Ads.manager(function(_, $)
     this.loadMetadata = function(adManagerMetadata, backlotBaseMetadata, movieMetadata)
     {
       this.ready = true;
+
+      if (adManagerMetadata)
+      {
+        if (adManagerMetadata["cacheBuster"] === "false")
```
pilievOoyala: in case we change the default, can you add in reading if === "true"

pilievOoyala:
LGTM

PR #126, opened 2016-07-22
PBI-1828  In some cases VPAID ads can have multiple "MediaFile" tags …

pilievOoyala:
…in the vast response.  Adding it code to just grab the first one. That way, when you grab the url, it doesn't just append all the urls together.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -2650,7 +2650,9 @@ OO.Ads.manager(function(_, $) {
       }
 
       var $paramsNode = $node.find('AdParameters');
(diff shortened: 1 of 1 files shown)


pilievOoyala:
Please double check that this is a valid solution.  I don't see anything in the vpaid spec other than that there should be only 1 tag, but IMA parses this without a problem, so we should fix this issue.

Here is the test page:
http://goo.gl/LEjtbK

dchauOoyala:
Seems like a valid solution to me

dchauOoyala:
LGTM

PR #128, opened 2016-07-27
pbi-1873  Changing the way that SSAI appends params to a url.  It now…

pilievOoyala:
… checks to see if the ? exists and appends the params correctly now.

### DIFF
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -578,7 +578,7 @@ OO.Ads.manager(function(_, $)
       var newUrl = _appendParamToUrl(url, offset);
 
       var adIdParam = AD_ID_PARAM + adId;
-      newUrl = _appendParamToUrl(url, adIdParam);
+      newUrl = _appendParamToUrl(newUrl, adIdParam);
       return newUrl;
     }, this);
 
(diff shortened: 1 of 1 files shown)


js/ssai_pulse.js
```diff
@@ -574,11 +574,37 @@ OO.Ads.manager(function(_, $)
      */
     var _appendAdsProxyQueryParameters = _.bind(function(url, adId)
     {
-      // vastUrl + '&offset=5&aid=' + adid
-      return url + OFFSET_PARAM + OFFSET_VALUE + AD_ID_PARAM + adId;
+      var offset = OFFSET_PARAM + OFFSET_VALUE;
+      var newUrl = _appendParamToUrl(url, offset);
+
+      var adIdParam = AD_ID_PARAM + adId;
+      newUrl = _appendParamToUrl(url, adIdParam);
```
dchauOoyala: I think appendParamToUrl should be taking in the newUrl here, as opposed to url.

js/ssai_pulse.js
```diff
@@ -574,11 +574,37 @@ OO.Ads.manager(function(_, $)
...
+        if (url.indexOf("?") > -1)
+        {
+          return url + "&" + param;
+        }
+        else {
+          return url + "?" + param;
+        }
+      }
+      return "";
+    }
```
dchauOoyala: semicolon

PR #129, opened 2016-07-27
Moat integration and initialization

cescamilla:
add the integration of Moat for Meredith and future customers who wish this service, i used google_ima.js file as base and added Moat function and initialise it when Google IMA object is initialised.


pilievOoyala:
This is a custom integration of an analytics module into an ad module. So we can't check this into our repo.

PR #132, opened 2016-08-12
[google-ima-debug-flag] Putting in correct flag check for v4 so Googl…

pilievOoyala:
…eIMA plugin will load the debug sdk when debug = true (i.e. our debug page).


aeng7:
LGTM

PR #133, opened 2016-08-15
PSDK-520: Fix timeout bug on iOS due to wrong expectations of the plu…

apesant:
…gin on the order of the events coming from the player

@pilievOoyala Can you review this? The timing with which we received the playAd call was  different on iOS which caused one of our variables (the pod ID we used to tell we started a pod) to be null in some cases. This commit fixes that.

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -47,7 +47,6 @@
              */
             this.initialize = function(adManagerController, playerId) {
                 amc = adManagerController; // the AMC is how the code interacts with the player
-                amc.adManagerWillControlAds();
                 pulseAdManagers[playerId] = this;
 
                 // Add any player event listeners now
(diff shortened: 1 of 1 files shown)


js/pulse.js
```diff
@@ -47,6 +47,7 @@
              */
             this.initialize = function(adManagerController, playerId) {
                 amc = adManagerController; // the AMC is how the code interacts with the player
+                amc.adManagerWillControlAds();
```
pilievOoyala: Have you tested replaying the same video or watching videos back to back with this?  I don't think initialize will get run again. So the adManagerWillControlAds won't work the second time around.  If you would like an example, please refer to google_ima.js.  Inside that ad manager the adManagerWillControlAds is called when the ad manager tries to do the ad request.  That way, if you don't have an ad tag, you won't mess with the flow of the player.

apesant:
@pilievOoyala Thanks for the feedback, I updated the code to call adManagerWillControlAds when the initial player is requested. Is it ok now?

pilievOoyala:
LGTM now.  I'm assuming this is all tested on your side.

PR #134, opened 2016-08-24
PSDK-527: Add optional callbacks for emitting session & ad player

eshaglund:
Add the ability to pass callbacks, as page level parameters, which will receive internal session & ad player objects as they are created

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -37,6 +37,9 @@
             var amc  = null;
             var pulseSDKUrl = "/proxy/pulse-sdk-html5/2.1/latest.min.js";
             var adModuleJsReady = false;
+            var pluginCallbacks = {
+
+            };
 
             /**
              * Ad manager init
@@ -282,7 +285,17 @@
                     }
                 }
 
+                // Check for some callbacks to expose plugin internals
+                if(adManagerMetadata.pulse_callbacks) {
+                    var callback;
+                    for(var name in adManagerMetadata.pulse_callbacks) {
+                        callback = adManagerMetadata.pulse_callbacks[name];
 
+                        if(typeof callback === 'function') {
+                            pluginCallbacks[name] = callback;
+                        }
+                    }
+                }
 
                 //The request settings and content metadata are going to be assembled progressively here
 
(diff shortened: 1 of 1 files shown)


eshaglund:
@pilievOoyala would be nice to have someone look over this. It's what we came up with for exposing some plugin internals when required by client implementations

js/pulse.js
```diff
@@ -282,7 +285,17 @@
                     }
                 }
 
+                // Check for some callbacks to expose plugin internals
```
pilievOoyala: When a new embed_code (movie) is set, we run through this code again.  If the client doesn't want callbacks the second time around, it would be good to clear out all the old callbacks before parsing the new ones.

pilievOoyala:
In general can you jsdoc which callbacks are supported, so the syntax for the key value is clear.

js/pulse.js
```diff
@@ -643,6 +656,10 @@
                         adPlayer.addEventListener(OO.Pulse.AdPlayer.Events.AD_CLICKED, _.bind(_onAdClicked,this));
                         adPlayer.addEventListener(OO.Pulse.AdPlayer.Events.LINEAR_AD_PAUSED, _.bind(_onAdPaused,this));
                         adPlayer.addEventListener(OO.Pulse.AdPlayer.Events.LINEAR_AD_PLAYING, _.bind(_onAdPlaying,this));
+
+                        if(pluginCallbacks.onAdPlayerCreated) {
```
pilievOoyala: Please add an extra null check for pluginCallbacks. You may guarantee it's not null now, but if someone changes the code in the future, it may break during this if check.

js/pulse.js
```diff
@@ -659,6 +676,9 @@
                         // //console.log('> ' + logItem.source + ': ' + logItem.message);
                     });
                     session = OO.Pulse.createSession(this._contentMetadata, this._requestSettings);
+                    if(pluginCallbacks.onSessionCreated) {
```
pilievOoyala: Same here, extra null check for pluginCallbacks.

eshaglund:
Clearing the callbacks is a good point, will take care of that.

As for jsdoc on them, the rest of the page-level parameters are not documented in the code, but rather on support.ooyala.com, so I think it makes sense to describe them in a consistent manner.

Finally, I think a null check for the callback list is redundant as the value of pluginCallbacks is constant. If somebody changes this to actually be variable they should also update the logic handling it.

pilievOoyala:
I agree that if someone changes pluginCallbacks they "should" update the logic.  But more than likely, it will get missed or one of the instances will get missed.  Adding in the extra null check is a failsafe for the future and may avoid a ticket and wasted time debugging.

As far as the documenting, I now see we haven't done that for other plugins as well. We should try and change that in the future, but it's fine for now. (Since it's open source, it would be nice to have everything documented in one spot. Then people don't have to look at our support page to get all the info.)

eshaglund:
I guess it's not a huge deal, so I've changed it. Also removed some unused code.

pilievOoyala:
Looks good. Thank you!

PR #135, opened 2016-08-25
[PBI-1927]

aeng7:
-fixed an issue where the video would flash black first prior to rendering an IMA overlay


pilievOoyala:
lgtm.

PR #137, opened 2016-08-26
[PBW-5838]

aeng7:
-reverted IMA overlay blink changes because of side effects. Will revisit in the future.


pilievOoyala:
Looks good.

PR #139, opened 2016-09-08
[PBI-1951]

dchauOoyala:
- Add new AdManagerUtils class and unit tests for utility class
- Fix VMAP logic error caused by AdManagerUtils#convertTimeStampToMilliseconds


utils/ad_manager_utils.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return milliseconds;
+  };
+
+  /**
+   * Helper function to convert a percentage representing time into milliseconds.
+   * @private
+   * @method AdManagerUtils#_convertPercentToMilliseconds
+   * @param {string} timeString The string that represents a percentage (format: [0, 100]%)
+   * @param {number} movieDuration The duration of the video - represented in seconds
```
aeng7: Can we rename this parameter? The util as defined shouldn't have anything to do with movies or their durations. Something like totalDuration might work. We should also update the description of the jsdoc to include a bit about how this duration value is used.

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,103 @@
...
+  }, this);
+
+  /**
+   * Helper function to determine if the movie duration is valid.
+   * @private
+   * @method AdManagerUtils#_isValidMovieDuration
+   * @param {number} movieDuration The duration of the video - represented in seconds
+   * @returns {boolean} true if the movie duration string is valid. Returns false if otherwise.
+   */
+  var _isValidMovieDuration = _.bind(function(movieDuration)
```
aeng7: Can we try to remove the reference to "movie" in this util? This will make it a generic util.

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,103 @@
...
+      var minutesInSeconds = +hms[1] * 60;
+      var seconds = +hms[2];
+      milliseconds = (hoursInSeconds + minutesInSeconds + seconds) * 1000;
+    }
+    return milliseconds;
+  };
+
+  /**
+   * Helper function to convert a percentage representing time into milliseconds.
+   * @private
```
aeng7: Is this a private or public function?

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,103 @@
...
+ * @public
+ * @class AdManagerUtils
+ * @classdesc Provides utility/helper functions for ad managers.
+ */
+var AdManagerUtils = function()
+{
+
+  /**
+   * Helper function to convert the HMS timestamp into milliseconds.
+   * @private
```
aeng7: Is this a public or private function?

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,103 @@
...
+   * Helper function to determine if a time stamp string is valid.
+   * @private
+   * @method AdManagerUtils#_isValidTimeString
+   * @param {string} timeString The string that represents a percentage (format: [0, 100]%)
+   * @returns {boolean} true if the time stamp string is valid. Returns false if otherwise.
+   */
+  var _isValidTimeString = _.bind(function(timeString)
+  {
+    var result = true;
+    if (!timeString || !_.isString(timeString))
```
aeng7: Can we add a check to this function to test that it has the correct format (hh:mm:ss)? This will ensure that the references to hms[0], hms[1], and hms[2] above won't throw an exception.

utils/vast_parser.js
```diff
@@ -0,0 +1,533 @@
+var VastParser = function() {
```
aeng7: Were there any changes from OO.VastParser to this?

test/unit-tests/ad_manager_utils_test.js
```diff
@@ -0,0 +1,94 @@
...
+ * Unit test class for the Vast Ad Manager
+ * https://github.com/Automattic/expect.js
+ */
+
+var adManagerUtils = require("../../utils/ad_manager_utils.js");
+
+describe('Ad Manager Utility Class:', function()
+{
+
+  it('Time Stamps Should Be Converted to Milliseconds', function()
```
aeng7: Can we add unit tests with improper hh:mm:ss formats? Such as 10:10: or :10:10 or 10:10:10:10?

dchauOoyala: I originally did not add the unit tests for improper hh:mm:ss formats because the caller of the utility function (ad_manager_vast.js) does regex filtering for those improper formats. But yes, I'll add checks for improper hh:mm:ss to the function for the sake of this utility class being generic.

utils/ad_manager_utils.js
dchauOoyala: will do

utils/vast_parser.js
dchauOoyala: no other changes

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+var AdManagerUtils = function()
+{
+
+  /**
+   * Converts the hh:mm:ss timestamp into milliseconds.
+   * @public
+   * @method AdManagerUtils#convertTimeStampToMilliseconds
+   * @param {string} timeString The timestamp string (format: hh:mm:ss / hh:mm:ss.mmm)
+   * @returns {number||null} The number of milliseconds the timestamp represents. Returns null
```
aeng7: I think multiple return types use a single pipe rather than double pipe.

utils/ad_manager_utils.js
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  /**
+   * Helper function to validate that a string is (pseudo) valid hh:mm:ss format.
+   * @private
+   * @method AdManagerUtils#_isValidHms
+   * @param {string} hms The hh:mm:ss string
+   * @returns {boolean} true if the hh:mm:ss string is valid. Returns false if otherwise.
+   */
+  var _isValidHms = _.bind(function(hms)
```
aeng7: Very nice function!

aeng7:
LGTM other than a minor jsdoc issue

js/ad_manager_vast.js
```diff
@@ -2657,12 +2659,12 @@ OO.Ads.manager(function(_, $) {
             break;
           // case: hh:mm:ss.mmm | hh:mm:ss
           case /^\d{2}:\d{2}:\d{2}\.000$|^\d{2}:\d{2}:\d{2}$/.test(adBreak.timeOffset):
```
pilievOoyala: Since you're in here already.... can you change this to some if/else statements? I just have a fear that some browser is going to interpret switch statements differently and we'll end up with some bugs.

utils/ad_manager_utils.js
pilievOoyala: yes. high 5

pilievOoyala:
LGTM after you get teh other changes in.

test/unit-tests/ad_manager_utils_test.js
```diff
@@ -0,0 +1,106 @@
...
+    milliseconds = adManagerUtils.convertPercentToMilliseconds(null, 100);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertPercentToMilliseconds("100%", null);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertPercentToMilliseconds(false, null);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertPercentToMilliseconds(false, 50);
```
pilievOoyala: what about undefined?

pilievOoyala: or negative percents

test/unit-tests/ad_manager_utils_test.js
```diff
@@ -0,0 +1,106 @@
...
+var adManagerUtils = require("../../utils/ad_manager_utils.js");
+
+describe('Ad Manager Utility Class:', function()
+{
+
+  it('Time Stamps Should Be Converted to Milliseconds', function()
+  {
+    var milliseconds;
+
+    milliseconds = adManagerUtils.convertTimeStampToMilliseconds("00:00:00");
```
pilievOoyala: what about negative numbers for each position?

test/unit-tests/ad_manager_utils_test.js
```diff
@@ -0,0 +1,106 @@
...
+    milliseconds = adManagerUtils.convertTimeStampToMilliseconds(0);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertTimeStampToMilliseconds(false);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertTimeStampToMilliseconds(null);
+    expect(milliseconds).to.be(null);
+
+    milliseconds = adManagerUtils.convertTimeStampToMilliseconds(undefined);
```
pilievOoyala: also, what happens if you pass an empty object?

test/unit-tests/ad_manager_utils_test.js
pilievOoyala: something that isn't a percent in the string, and pass in an empty object instead of a string.

test/unit-tests/ad_manager_utils_test.js
pilievOoyala: also what about absurdly large numbers?

dchauOoyala: Added check for negative numbers. Any argument without 2 ":"'s will not be accepted. As for large numbers, that would be a limitation of the language specification, which in Javascript's case is: 9007199254740991

pilievOoyala:
LGTM.

PR #143, opened 2016-09-20
[raise-timeout] Raising IMA's ad requesting time-out to 5 seconds.

pilievOoyala:
We are raising the timeout because VPAID 1.0 ads seem to take longer to load and this will hopefully eliminate some of the customer headache of losing revenue from ads failing to load in time.


dchauOoyala:
lgtm

aeng7:
LGTM

PR #145, opened 2016-09-21
[PBW-5938]

aeng7:
-fixed an issue where a linear video playing in single video element mode would not end the current ad pod when it finishes
-removed outdated comments
-fixed an issue in the ima unit tests where the amc was not a new object at the beginning of each test

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1821,14 +1821,12 @@ require("../html5-common/js/utils/utils.js");
             }
             else
             {
-              //End the fake ad here
               _endCurrentAdPod(this.currentAMCAdPod.isRequest);
             }
           }
           else
           {
-            //End the fake ad here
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -1821,14 +1821,12 @@ require("../html5-common/js/utils/utils.js");
...
             {
-              //End the fake ad here
               _endCurrentAdPod(this.currentAMCAdPod.isRequest);
             }
           }
           else
           {
-            //End the fake ad here
-            _endCurrentAdPod(this.currentAMCAdPod.isRequest);
+            _endCurrentAdPod(true);
```
dchauOoyala: what about if the ad pod is an overlay?

aeng7: If the ad is an overlay, it'll go through the flow above. The currentIMAAd object will exist.

pilievOoyala:
looks good!

PR #150, opened 2016-10-06
[PBI-1958]

dchauOoyala:
Remove second -> millisecond conversion of AD_LOAD_TIMEOUT constant because it is already converted


pilievOoyala:
If that number were bad it should have caused what felt like infinite waiting?

pilievOoyala:
We talked it out in person. LGTM.

PR #151, opened 2016-10-11
[PBI-1995]

aeng7:
-changed AD_REQUEST ADTYPE to UNKNOWN_AD_REQUEST ADTYPE so that the ad video element gets created


apesant:
@aeng7 LGTM!

PR #152, opened 2016-10-19
Pbi 2031

aeng7:
-fixed an issue in IMA where invalid timeout values from the AMC were causing ads to timeout immediately


pilievOoyala:
excellent work!

PR #154, opened 2016-10-26
Pbi 2015

aeng7:
-Vast page level settings
Added support for all_ads array of ad objects. An ad object should have a tag_url, position_type, and position. This was done to match IMA page level ad support: http://help.ooyala.com/video-platform/concepts/pbv4_ads_dev_google_ima_parameters.html


pilievOoyala:
Looks great.  Just want to double check about page level ad frequency settings if possible. Otherwise that can go in a separate checkin.

aeng7:
The parameters "frequency" and "first_shown" can be added to each ad object in the all_ads array. This is not currently documented though. Works for Vast and IMA. Freewheel has its own page level/backlot format for settings, and does not currently support individual ad definitions.

PR #155, opened 2016-10-27
[PBI-2015]

aeng7:
-added back in ad tag url overrides from page level settings
Doc for reference: http://help.ooyala.com/video-platform/concepts/pbv4_ads_dev_vast_parameters.html


pilievOoyala:
Looks good

PR #157, opened 2016-11-02
[PBI-2051] Fix issue where closing overlay ad resets playhead to 0

dchauOoyala:
- only update play head on COMPLETE (IMA event) if the ad linear

### DIFF
--- a/test/unit-test-helpers/mock_ima.js
+++ b/test/unit-test-helpers/mock_ima.js
@@ -167,7 +167,7 @@ google =
       {
         ALL_ADS_COMPLETED : "allAdsCompleted",
         COMPLETE : "complete",
-        SKIPPED : "skipped",
+        SKIPPED : "skip",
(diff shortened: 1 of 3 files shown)


test/unit-test-helpers/mock_ima.js
```diff
@@ -167,7 +167,7 @@ google =
       {
         ALL_ADS_COMPLETED : "allAdsCompleted",
         COMPLETE : "complete",
-        SKIPPED : "skipped",
+        SKIPPED : "skip",
```
pilievOoyala: Is this right?

dchauOoyala: Yes, it actually wasn't getting caught originally because we have "skipped", but it's really supposed to be "skip"

pilievOoyala:
lgtm.
Also merginating

PR #159, opened 2016-11-07
[player-54] Fixing bad merge from before

pilievOoyala:


pilievOoyala:
I don't know if this is old or not. Closing. We'll make a new one.

PR #167, opened 2016-11-16
[PBF-731]  Pulled down update to html5-common

brianfwong:


brianfwong:
This seems to have picked up too many things. Closing and taking another look.

PR #177, opened 2017-01-24
[fixGoogleIMAStyleIfCheck] Fixing minification error. It was strippin…

pilievOoyala:
…g the if statement.


aeng7:
When does this situation happen when the iframe and its style does not exist? If it has not initialized yet, we should consider finding a definitive location in the code to set the z index. Otherwise the zIndex would not get applied, which would affect a few of our customers.

aeng7:
LGTM

PR #178, opened 2017-01-25
PSDK-762: Don't stop ad pod just because SDK doesn't load near-instantly

eshaglund:
Additionally, allow enabling Pulse debug mode by passing `pulse_debug` parameter to ad manager.

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -1,15 +1,29 @@
 /*
  * Pulse ad player ad manager
- *
- * version 1.0
  */
 
 (function(_, $)
 {
-
     var pulseAdManagers = {};
 
     OO.Ads.manager(function(_, $) {
+        function log() {
+            var args = Array.prototype.slice.call(arguments);
+            if(OO.Pulse) {
+                OO.Pulse.Utils.log.apply(null, args);
+            } else {
+                args.unshift('OO.Pulse: ');
+                console.log.apply(window.console, args);
+            }
+        }
+
+        var AD_MODULE_STATE = {
+            Uninitialized: 'uninitialized',
+            Loading: 'loading',
+            Ready: 'ready',
+            Failed: 'failed'
+        };
+
         /**
          * @class PulseAdManager
          * @classDesc The Pulse Ad Manager class.
@@ -36,7 +50,8 @@
             var preferredRenderingMode = null;
             var amc  = null;
             var pulseSDKUrl = "/proxy/pulse-sdk-html5/2.1/latest.min.js";
-            var adModuleJsReady = false;
+            var adModuleState = AD_MODULE_STATE.Uninitialized;
+            var enableDebugMode = false;
             var pluginCallbacks = {
 
             };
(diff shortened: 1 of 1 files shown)


js/pulse.js
```diff
@@ -1,15 +1,29 @@
...
+            if(OO.Pulse) {
+                OO.Pulse.Utils.log.apply(null, args);
+            } else {
+                args.unshift('OO.Pulse: ');
+                console.log.apply(window.console, args);
+            }
+        }
+
+        var AD_MODULE_STATE = {
+            Uninitialized: 'uninitialized',
```
dchauOoyala: Just a nitpicky comment: since this is "constants" object, perhaps the properties should also be all caps
(i.e. `UNINITIALIZED` as opposed to `Uninitialized`)

js/pulse.js
```diff
@@ -610,8 +650,9 @@
             };
 
             this.illegalOperationOccurred = function(msg) {
-                //console.log(msg);
+
```
dchauOoyala: If this function does nothing then should it be removed?

js/pulse.js
```diff
@@ -272,16 +287,22 @@
...
 
-                //Load the Pulse SDK if not already included
+                // Load the Pulse SDK if not already included
                 if(!OO.Pulse){
+                    adModuleState = AD_MODULE_STATE.Loading;
                     amc.loadAdModule(this.name, protocol + pulse_account_name + pulseSDKUrl, _.bind(function(success) {
-                        adModuleJsReady = success;
-                        if(isWaitingForPrerolls){
+                        adModuleState = success ? AD_MODULE_STATE.Ready : AD_MODULE_STATE.Failed;
+                        if(OO.Pulse)
```
dchauOoyala: We should definitely wrap this check with the proper braces `{}`. @pilievOoyala recently discovered that the minifier in the build process omits this style of code ( a conditional with the next line of code assumed as the branch) in the final minified file. This is typically interpreted correctly but there is possibly a bug in the minifier, so to be safe we should add the proper braces here.

js/pulse.js
```diff
@@ -673,10 +714,13 @@
             this.tryInitAdPlayer = function(){
                 var flashVersion = getFlashVersion().split(',').shift();
 
-                if(this.ui && adModuleJsReady) {
-
+                if(this.ui && adModuleState == AD_MODULE_STATE.Ready) {
```
dchauOoyala: We should prefer the use of `===` over `==` as this will avoid any unwanted implicit conversion that the interpreter may perform.

js/pulse.js
```diff
@@ -707,15 +751,15 @@
             var _onInitialPlay = function() {
                 isWaitingForPrerolls = true;
                 amc.adManagerWillControlAds();
-                if(adModuleJsReady){
+                if(adModuleState == AD_MODULE_STATE.Ready) {
```
dchauOoyala: Same comment as above about `==`

dchauOoyala:
Left a few comments, but not a deal breaker.

js/pulse.js
eshaglund: I like this style, but it's no big deal, I'll change it.

js/pulse.js
eshaglund: It's currently a required callback on the ad player listener interface

js/pulse.js
eshaglund: Oops, this change shouldn't even be in here. Good catch!

js/pulse.js
eshaglund: Mistake. Been writing too much Lua lately I think ..

PR #181, opened 2017-02-02
[PLAYER-551] Set browser locale setting for google ima

dchauOoyala:


js/google_ima.js
```diff
@@ -1034,6 +1034,7 @@ require("../html5-common/js/utils/utils.js");
         //These are required by Google for tracking purposes.
         google.ima.settings.setPlayerVersion(PLUGIN_VERSION);
         google.ima.settings.setPlayerType(PLAYER_TYPE);
+        google.ima.settings.setLocale(OO.getLocale());
```
aeng7: Do we need to do any validation before setting the locale? Like if getLocale() returns null/something unexpected.

dchauOoyala: getLocale(): `return (OO.locale || document.documentElement.lang || navigator.language || navigator.userLanguage || "en").substr(0,2).toUpperCase();`

The getLocale will default to `"EN"` if everything before it is falsey.

aeng7: The page https://www.loc.gov/standards/iso639-2/php/English_list.php that is linked to from the setLocale API at https://developers.google.com/interactive-media-ads/docs/sdks/html5/v3/apis#ima.AdsManager shows using "en" as lower case. Will "EN" still work?

dchauOoyala: Yes, I tested what would happen if capital letters were passed to IMA; it works. It looks like they probably lowercase the input to setLocale().

image

aeng7:
LGTM!

PR #182, opened 2017-02-08
[PLAYER-319] - (IMA) Added support for non-linear ads when using Ad Rules or VMAP

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-319)

The cause of the issue was similar to what was described on the ticket. The IMA sdk was trying to initialize the overlay, but since the _AdDisplayContainer_ hadn't been set up yet, it bailed out with the following error:

_AdError 501: The overlay ad content could not be displayed since creative dimensions do not align with display area._

With these changes we get notified right before the ad is about to be displayed, so I'm using this cue to initialize the _AdDisplayContainer_. This also hooks up to the AMC so that the `Ad Impression` event gets fired. These changes also fix the exact same issue when using VMAP.

**Known issue:**
The IMA guys confirmed that there is a bug in the SDK that prevents the `COMPLETE` event from getting fired when the overlay is removed. This means that non-linear ads that have a `minSuggestedDuration` will leave behind the transparent ad container when the ad is removed. This will block clicks in a portion of the player.

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1095,11 +1095,8 @@ require("../html5-common/js/utils/utils.js");
         _IMA_SDK_destroyAdsLoader();
         _IMAAdsLoader = new google.ima.AdsLoader(_IMAAdDisplayContainer);
         // This will enable notifications whenever ad rules or VMAP ads are scheduled
-        // for playback. Only needed on ad rules mode
-        if (_usingAdRules)
-        {
-          _IMAAdsLoader.getSettings().setAutoPlayAdBreaks(false);
-        }
+        // for playback, it has no effect on regular ads
+        _IMAAdsLoader.getSettings().setAutoPlayAdBreaks(false);
         _IMAAdsLoader.addEventListener(adsManagerEvents.ADS_MANAGER_LOADED, _onAdRequestSuccess, false);
         _IMAAdsLoader.addEventListener(adErrorEvent.AD_ERROR, _onImaAdError, false);
       });
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1202,7 +1202,6 @@ require("../html5-common/js/utils/utils.js");
       {
         if(_usingAdRules)
         {
-          _resetUIContainerStyle();
           //if ads are not ready yet, ima failed to load
           if(!this.adsReady)
           {
@@ -1942,6 +1941,7 @@ require("../html5-common/js/utils/utils.js");
           }
         }
 
+        _resetUIContainerStyle();
         this.currentIMAAd = null;
         this.adPlaybackStarted = false;
       });
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -47,6 +47,7 @@ require("../html5-common/js/utils/utils.js");
       var _linearAdIsPlaying;
       var _timeUpdater = null;
       var _uiContainer = null;
+      var _uiContainerPrevStyle = null;
 
       //Constants
       var DEFAULT_IMA_IFRAME_Z_INDEX = 10004;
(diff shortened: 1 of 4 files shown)


js/google_ima.js
```diff
@@ -1316,6 +1325,41 @@ require("../html5-common/js/utils/utils.js");
...
+        {
+          this.cancelAd(this.currentAMCAdPod);
+        }
+        // [PLAYER-319]
+        // IMA will not initialize ad rules overlays unless the ad container is already rendered and
+        // has enough room for the overlay by the time the ad is ready to play. As a workaround, we expand
+        // the ad container and make sure it's rendered, while at the same time hiding it visually.
+        // We store the element's current style in order to restore it afterwards.
+        _uiContainerPrevStyle = _uiContainer.getAttribute("style") || "";
+        _uiContainer.setAttribute("style", "display: block; width: 100%; height: 100%; visibility: hidden; pointer-events: none;");
```
jpfranco: I considered adding a message bus event for this, but I didn't want to expose functionality that is only useful for an IMA workaround. This is a very specific IMA restriction, so I think the skin doesn't necessarily have to be aware of it.

jpfranco:
Found a bug, closing until I'm able to fix it.

js/google_ima.js
```diff
@@ -1093,6 +1094,9 @@ require("../html5-common/js/utils/utils.js");
         var adErrorEvent = google.ima.AdErrorEvent.Type;
         _IMA_SDK_destroyAdsLoader();
         _IMAAdsLoader = new google.ima.AdsLoader(_IMAAdDisplayContainer);
+        // This will enable notifications whenever ad rules or VMAP ads are scheduled
+        // for playback, it has no effect on regular ads
+        _IMAAdsLoader.getSettings().setAutoPlayAdBreaks(false);
```
jpfranco: We need to disable `autoPlayAdBreaks` even if we're not currently using ad rules because this code is only executed once per each set of videos and we might have to handle an ad rules tag in the future. This setting doesn't have any effect on regular ads.

pilievOoyala:
LGTM but @aeng7  can you review as well?

test/unit-tests/ima_test.js
```diff
@@ -417,6 +417,55 @@ describe('ad_manager_ima', function()
...
+    initAndPlay(true, vci);
+    var am = google.ima.adManagerInstance;
+    am.publishEvent(google.ima.AdEvent.Type.AD_BREAK_READY);
+    am.publishEvent(google.ima.AdEvent.Type.LOADED);
+    am.publishEvent(google.ima.AdEvent.Type.STARTED);
+    // These should exist when overlay is being displayed
+    expect(ima.currentAMCAdPod).to.be.ok();
+    expect(ima.currentIMAAd).to.be.ok();
+    am.publishEvent(google.ima.AdEvent.Type.AD_BREAK_READY);
+    // These should be removed when next ad break is ready
```
aeng7: Should these be removed when the current ad break is over rather than the next ad break is ready?

aeng7:
Should we also add _resetUIContainerStyle(); to this.cancelAd or _endCurrentAd. It's possible for the ad to get cancelled from outside the plugin (such as if the user clicks to a discovery video right as the ad is about to play).

test/unit-tests/ima_test.js
jpfranco: Yes, this is mostly because there is no `COMPLETE` event for non-linear ads, so there is no way to know when the ad is removed. It's also a prerequisite for this fix that the current ad gets removed before the next ad begins because we need to have `_uiContainer` in its default state.

jpfranco:
@aeng7 hmm I'll have to double check. The `_uiContainer` style is reset on the `LOADED` event, so I'm not sure if it's possible to cancel an ad before this event fires. If it's possible then we do need to add a call to `_resetUIContainerStyle()`.

jpfranco:
@aeng7 I moved the `_resetUIContainerStyle()` call to `_endCurrentAd`, this way all cases will be covered (ad errors, cancellations, etc.).

aeng7:
LGTM!

PR #183, opened 2017-02-09
[fixIMAUnitTests] Unit tests were broken when we adding locale settin…

pilievOoyala:
…gs to ima. Fixing the unit tests because they were trying to access a variable in OO that wasn't defined for tests.


dchauOoyala:
LGTM

PR #184, opened 2017-02-09
[player-552] Freewheel now interprets correctly the fw_video_asset_ne…

pilievOoyala:
…twork_id param coming from backlot.

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -129,7 +129,7 @@ OO.Ads.manager(function(_, $) {
         }
         profileId = metadata['html5_player_profile'];
         siteSectionId = metadata['fw_site_section_id'];
-        videoAssetId = metadata['fw_video_asset_id'] || videoAssetId;
+        videoAssetId = metadata['fw_video_asset_id'] || metadata['fw_video_asset_network_id'] || videoAssetId;
         FRMSegment = metadata['FRMSegment'];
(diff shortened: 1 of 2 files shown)


js/freewheel.js
```diff
@@ -146,12 +146,15 @@ OO.Ads.manager(function(_, $) {
       // [PBW-449] To maintain parity with Flash, we allow 'fw_video_asset_id'
       // to be set on the page-level (see 'onPlayerCreated') or fallback to
       // movie metadata, then externalId, and then embed_code.
```
aeng7: Does this comment still apply without modifications?

pilievOoyala: Yes, I just simplified the logic. It does the same thing.  The only real change is the one above.

test/unit-tests/freewheel_test.js
```diff
@@ -150,6 +214,30 @@ describe('ad_manager_freewheel', function() {
     expect(videoAssetId).to.be("myEmbedCode");
   });
 
+  it('Init: test video asset override part 4', function(){
```
aeng7: Can we add more details to the unit test titles? Rather than part x, can we include what each test is testing (such as when fw_video_asset_id has priority, etc).

aeng7:
LGTM

PR #185, opened 2017-02-09
[PBI-2100]

aeng7:
-fixed an issue in the vast ad manager where wrapper ads would cause the ad pod to end early (before wrapper ads are processed). This caused a content flicker for preroll ads


pilievOoyala:
LGTM

PR #186, opened 2017-02-16
Player 552 b

pilievOoyala:


js/google_ima.js
```diff
@@ -279,6 +279,12 @@ require("../html5-common/js/utils/utils.js");
           this.useInsecureVpaidMode = metadata.vpaidMode === "insecure";
         }
 
+        this.disableFlashAds = false;
```
aeng7: Should we also define this and set to false in the _resetVars function? The other page level settings are treated like this.

test/unit-test-helpers/mock_ima.js
```diff
@@ -64,7 +64,8 @@ google =
       setPlayerVersion : function() {},
       setPlayerType : function() {},
       setVpaidMode : function() {},
-      setLocale : function() {}
+      setLocale : function() {},
+      setDisableFlashAds : function() {},
```
aeng7: Can we add a unit test to check that the page level setting reaches this mock setDisableFlashAds function with the correct value?

js/google_ima.js
pilievOoyala: I totally forgot about that. good catch.

test/unit-tests/ima_test.js
```diff
@@ -1512,4 +1512,28 @@ describe('ad_manager_ima', function()
     videoWrapper.destroy();
     expect(ima.sharedVideoElement).to.be(null);
   });
+
+  it('Test disabling flash ads flag', function()
+  {
+    var flag;
+    google.ima.settings.setDisableFlashAds = function(value) {flag = value};
```
pilievOoyala: I need to remove this. I rushed this test out.

test/unit-test-helpers/mock_ima.js
pilievOoyala: Added below.  I dont think I can run the actual ad request code, where we set the param in the ima sdk but I can test that loadmetadata is reading it in correctly.

aeng7:
LGTM

PR #187, opened 2017-02-20
Prevent player from consuming pointer events while VPAID ads are displayed

eshaglund:


js/pulse.js
```diff
@@ -916,15 +965,17 @@
...
-            var _onAdFinished = function(){
+            var _onAdFinished = function() {
                 amc.notifyLinearAdEnded(1);
+                enableAdScreenPointerEvents();
             };
 
-            var _onAdSkipped = function(){
+            var _onAdSkipped = function() {
                 amc.notifyLinearAdEnded(1);
+                enableAdScreenPointerEvents();
```
aeng7: Is there a similar callback for when an ad errors out? If so, should we also enableAdScreenPointerEvents there as well?

js/pulse.js
```diff
@@ -252,46 +257,89 @@
...
+                if(values && values.length === 2) {
                     return values[1];
                 }
             }
 
+            var updateAdScreenPointerEventsEnabled = (function() {
+                var adScreens = document.getElementsByClassName('oo-ad-screen');
+                var skinClickLayers = document.getElementsByClassName('oo-player-skin-plugins-click-layer');
+
+                if(adScreens.length === 0) {
```
aeng7: Can we add a null check for both adScreens and skinClickLayers? This can happen if the customer decides to use their own skin, or otherwise not use our skin. Also, will we face the same issue if the customer implements a custom skin in the same manner? We might want to consider adding a hook for both the ad plugins and skin to sync up on skin layer clickability.

eshaglund: `getElementsByClassName` will return an empty `HTMLCollection` if no matching elements are found, so it should be fine.

js/pulse.js
eshaglund: Quite possibly, I'll add it if there is

eshaglund: I've added it, but without `notifyLinearAdEnded`.
Not every `LINEAR_AD_ERROR` has a corresponding `LINEAR_AD_START` so I'm not sure what the implications might be of calling `notifyLinearAdEnded` too often.

Since it's worked so far, I decided to just leave it out until we can look into it more.

js/pulse.js
aeng7: Sounds good

aeng7:
LGTM

PR #192, opened 2017-04-03
PSDK-840: Fix ad playback

eshaglund:
- Only add pause insertion point filter if filter is already present.

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -172,24 +172,25 @@
                 var POSTROLL = 4;
 
                 var pos = parseInt(position);
-                var insertPointFiler = [];
-
-                //Always add pause ads
-
-                insertPointFiler.push("onPause");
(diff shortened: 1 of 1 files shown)


js/pulse.js
```diff
@@ -172,24 +172,25 @@
                 var POSTROLL = 4;
 
                 var pos = parseInt(position);
-                var insertPointFiler = [];
-
-                //Always add pause ads
-
-                insertPointFiler.push("onPause");
+                var insertionPointFilter = [ ];
```
aeng7: Minor thing, but there is an extra space in this array declaration.

dchauOoyala:
LGTM

js/pulse.js
eshaglund: Actually, it is there on purpose

PR #194, opened 2017-04-24
set persistent id to the player device id

mananchandra:
This change is to allow the player to set the Pulse persistent ID so that it matches the device ID used in the video analytics SDK. This will allow us to get analytics on all the video and ad events for a given device.

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -908,7 +908,9 @@
             };
 
             var _onDeviceIdSet = function(event, deviceId) {
-                this._persistentId = deviceId
+                if (!this._persistentId) {
+                  this._persistentId = deviceId;
+                }
             };
 
             var _onFullscreenChanged = function(event, shouldEnterFullscreen)
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -91,6 +91,7 @@
                 amc.addPlayerListener(amc.EVENTS.SIZE_CHANGED, _.bind(_onSizeChanged, this));
                 amc.addPlayerListener(amc.EVENTS.FULLSCREEN_CHANGED, _.bind(_onFullscreenChanged, this));
                 amc.addPlayerListener(amc.EVENTS.REPLAY_REQUESTED, _.bind(_onReplay, this));
+                amc.addPlayerListener(amc.EVENTS.DEVICE_ID_SET, _.bind(_onDeviceIdSet, this));
             };
 
             this.getAdPlayer = function() {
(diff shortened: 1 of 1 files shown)


mananchandra:
@eshaglund @pilievOoyala can you please review this PR?

pilievOoyala:
Is the message already added to html5_common? If so, did you already update the subrepo for this repo?

js/pulse.js
```diff
@@ -904,6 +907,10 @@
                 }
             };
 
+            var _onDeviceIdSet = function(event, deviceId) {
+                this._persistentId = deviceId
```
eshaglund: missing semicolon

js/pulse.js
```diff
@@ -363,7 +364,9 @@
 
                 this.ready = true;
                 this._deviceContainer = adManagerMetadata.pulse_device_container;
-                this._persistentId = adManagerMetadata.pulse_persistent_id;
+                if (adManagerMetadata.pulse_persistent_id) {
+                  this._persistentId = adManagerMetadata.pulse_persistent_id;
```
eshaglund: This will not change it if already present in the metadata, but `_onDeviceIdSet` doesn't have such a check. Shouldn't it be there as well? Or how is `DEVICE_ID_SET` triggered?

mananchandra:
@pilievOoyala there is an existing event raised by the player called OO.EVENTS.GUID_SET. Therefore I did not need to change html5_common to make this work. Please take a look at the corresponding Mjolnir changes at https://git.corp.ooyala.com/projects/IO/repos/mjolnir/pull-requests/1/diff and let me know if something needs to change.

js/pulse.js
mananchandra: We want to use the device id being used by the player as the persistent id, except when the pulse_persistent_id is already defined in the metadata. In that case we will allow overriding the player device id, this is to maintain backward compatibility with the way it currently works.

DEVICE_ID_SET is triggered when the player identifies the device id from the local storage. This device id has been around for many years and used in identifying devices for video analytics. We want to use the same device id for ads.
The corresponding changes to the core player are in this PR that I have added you as a reviewer to https://git.corp.ooyala.com/projects/IO/repos/mjolnir/pull-requests/1/diff

DEVICE_ID_SET is triggered on loading the player therefore it gets triggered before loadMetadata

js/pulse.js
pilievOoyala: If _onDeviceIdSet is called after loadMetadata (I'm not sure if that's possible now) this would overwrite the _persistentId.  And I think the pulse_persistent_id is supposed to win if you have both specified.

mananchandra: I will add the check here. We still need the check I added in loadMetadata because otherwise the persistend id will be set to null if it did not exist in the metadata.

pilievOoyala:
LGTM.

PR #195, opened 2017-05-30
Player 1033

aeng7:
-fixed an issue where IMA ads can still load if the IMA plugin times out after sending out the ad request. This causes the player to go into a bad state

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -742,6 +742,7 @@ require("../html5-common/js/utils/utils.js");
         _resetAdsState();
         _resetPlayheadTracker();
         this.contentEnded = false;
+        this.adRulesLoadError = false;
         //In the case of ad rules, non of the ads are in the timeline
         //and we won't call initialPlayRequested again. So we manually call
         //to load the ads again. We don't care about preloading at this point.
@@ -749,7 +750,6 @@ require("../html5-common/js/utils/utils.js");
         {
           _trySetupAdsRequest();
         }
-        this.adRulesLoadError = false;
       });
 
       /**
@@ -1225,7 +1225,7 @@ require("../html5-common/js/utils/utils.js");
        */
       var _adsRequestTimeout = privateMember(function()
       {
-        OO.log("IMA Ad request timeout");
+        OO.log("IMA Ad request timed out");
         if (!this.adsReady)
         {
           _onImaAdError(OOYALA_IMA_PLUGIN_TIMEOUT);
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -65,6 +65,7 @@ require("../html5-common/js/utils/utils.js");
       var OVERLAY_HEIGHT_PADDING = 50;
 
       var TIME_UPDATER_INTERVAL = 500;
+      var OOYALA_IMA_PLUGIN_TIMEOUT = "ooyalaImaPluginTimeout";
 
       /**
        * Helper function to make functions private to GoogleIMA variable for consistency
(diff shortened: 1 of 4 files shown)


js/google_ima.js
```diff
@@ -725,6 +733,11 @@ require("../html5-common/js/utils/utils.js");
        */
       var _onReplayRequested = privateMember(function()
       {
+        if (!_IMAAdsLoader)
+        {
+          //The Ads Loader might have been destroyed if we had timed out.
+          IMA_SDK_tryCreateAdsLoader();
+        }
```
pilievOoyala: good catch

js/google_ima.js
```diff
@@ -65,6 +65,7 @@ require("../html5-common/js/utils/utils.js");
       var OVERLAY_HEIGHT_PADDING = 50;
 
       var TIME_UPDATER_INTERVAL = 500;
+      var OOYALA_IMA_PLUGIN_TIMEOUT = "ooyalaImaPluginTimeout";
```
pilievOoyala: I think this needs a better name

js/google_ima.js
```diff
@@ -736,6 +749,7 @@ require("../html5-common/js/utils/utils.js");
         {
           _trySetupAdsRequest();
         }
+        this.adRulesLoadError = false;
```
pilievOoyala: should this be before the trySetupAdRequest?

js/google_ima.js
```diff
@@ -1199,9 +1225,10 @@ require("../html5-common/js/utils/utils.js");
        */
       var _adsRequestTimeout = privateMember(function()
       {
+        OO.log("IMA Ad request timeout");
```
pilievOoyala: Can this be "IMA Ad request timed out"?

js/google_ima.js
pilievOoyala: Nevermind, I misread it and didn't see it was a string.

js/google_ima.js
```diff
@@ -299,6 +301,11 @@ require("../html5-common/js/utils/utils.js");
         {
           _IMA_SDK_tryInitAdContainer();
         }
+        else if (!_IMAAdsLoader)
```
jpfranco: Is this an `else if` because `_IMA_SDK_tryInitAdContainer` also creates the Ads Loader?

js/google_ima.js
```diff
@@ -1150,8 +1178,6 @@ require("../html5-common/js/utils/utils.js");
        */
       var _IMA_SDK_destroyAdsManager = privateMember(function()
       {
-        this.currentIMAAd = null;
-        this.currentNonLinearIMAAd = null;
```
jpfranco: Is this cleanup handled elsewhere now?

js/google_ima.js
```diff
@@ -1264,6 +1297,8 @@ require("../html5-common/js/utils/utils.js");
         {
           //destroy the current ad manager is there is one
           _IMA_SDK_destroyAdsManager();
+          this.currentIMAAd = null;
```
jpfranco: Does this mean that there are cases in which we want to destroy the AdsManager but also keep the current ads?

js/google_ima.js
aeng7: Yep, the AdsLoader is dependent on the Ad Container. The current flow is the create the Ads Loader after the Ad Container is created.

js/google_ima.js
aeng7: Yep, thats a better place to have it. Moving it there.

js/google_ima.js
aeng7: Yep, I've decoupled the IMAAd objects from the destruction of the ads manager. These two will be nulled if necessary at points after destroyAdsManager is called in the code.

js/google_ima.js
aeng7: Done.

js/google_ima.js
aeng7: We need to keep the currentIMAAd objects long enough to ensure a full cleanup. One such instance of this is in this.cancelAd, where we need these two for _endCurrentAd to cleanup properly.

pilievOoyala:
LGTM. Merging

PR #197, opened 2017-06-09
Pbw 6610

aeng7:
-fixed an issue where adsManager.start was not called immediately after adsManager.init
-changed flow of certain paths since adsManager.start is called immediately
-removed unneeded unit tests

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -147,7 +147,6 @@ require("../html5-common/js/utils/utils.js");
         this.hasPreroll = false;
 
         this.adPlaybackStarted = false;
-        this.vcPlayRequested = false;
         this.savedVolume = -1;
         this.showAdControls = false;
         this.useGoogleAdUI = false;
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -622,52 +621,18 @@ require("../html5-common/js/utils/utils.js");
...
-            this.adPlaybackStarted = true;
+            //resumeAd will only be called if we have exited fullscreen
+            //so this is safe to call
+            this.sharedVideoElement.webkitEnterFullscreen();
+            this.sharedVideoElement.play();
           }
-        }
-      };
-
-      this.requestPause = function()
```
pilievOoyala: Do we not need requestPause and requestPlay anymore?

js/google_ima.js
```diff
@@ -1582,6 +1547,7 @@ require("../html5-common/js/utils/utils.js");
           case eventType.USER_CLOSE:
           case eventType.SKIPPED:
           case eventType.COMPLETE:
+            this.adPlaybackStarted = false;
```
pilievOoyala: I think this needs to get added to the resetvars()

js/google_ima.js
aeng7: Nope, the video wrapper is calling pauseAd and resumeAd directly now.

js/google_ima.js
aeng7: It is already in resetVars.

PR #198, opened 2017-06-28
[PLAYER-1038]

aeng7:
-fixed an issue where the IMA ad plugin and video plugin consume IMA pause and resume events after receiving an IMA ad complete event. This would throw the player into a bad state

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1480,9 +1480,15 @@ require("../html5-common/js/utils/utils.js");
        * Callback from IMA SDK for ad tracking events.
        * @private
        * @method GoogleIMA#_IMA_SDK_onAdEvent
+       * @param {object} adEvent The IMA ad event
        */
       var _IMA_SDK_onAdEvent = privateMember(function(adEvent)
       {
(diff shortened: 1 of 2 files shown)


js/google_ima.js
```diff
@@ -1622,6 +1628,36 @@ require("../html5-common/js/utils/utils.js");
       });
 
       /**
+       * Checks to see if we should ignore certain IMA ad events if the ad
+       * is not currently playing (either completed or has not started).
+       * @private
+       * @method GoogleIMA#_ignoreWhenAdNotPlaying
+       * @param {object} adEvent The IMA ad event
+       */
+      var _ignoreWhenAdNotPlaying = privateMember(function(adEvent)
```
pilievOoyala: should we name this _checkIgnoreWhenAdNotPlaying?  I'm ok with what it is right now. Just want to see what you think.

aeng7: I think it's ok when looking at it from an "if (_ignoreWhenAdNotPlaying)" standpoint.

PR #200, opened 2017-07-12
[pulse-ready-fix]

aeng7:
Note that this is a suggestion and is not fully tested. Please feel free to make any changes necessary.

-fixed an issue where if there is no _pulseHost, the AMC has to wait until the Pulse ad plugin times out before notifying of playback ready

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -359,10 +359,14 @@
 
                 if(!this._pulseHost) {
                     log('No Pulse hostname found in plugin parameters or media metadata; will not attempt to show Pulse ads');
+                    this.ready = true;
+                    amc.onAdManagerReady();
                     return;
                 }
 
                 this.ready = true;
+                amc.onAdManagerReady();
+
                 this._deviceContainer = adManagerMetadata.pulse_device_container;
                 if (adManagerMetadata.pulse_persistent_id) {
                   this._persistentId = adManagerMetadata.pulse_persistent_id;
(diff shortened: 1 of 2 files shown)


js/pulse.js
```diff
@@ -359,10 +359,14 @@
 
                 if(!this._pulseHost) {
                     log('No Pulse hostname found in plugin parameters or media metadata; will not attempt to show Pulse ads');
+                    this.ready = true;
+                    amc.onAdManagerReady();
                     return;
                 }
 
                 this.ready = true;
+                amc.onAdManagerReady();
```
aeng7: Is it safe to call onAdManagerReady here? When we call this API, the AMC checks to see if all ad plugins are ready. If so, it unblocks playback ready. At this point, playback can occur when the user clicks the play icon or automatically if autoplay is on.

eshaglund: hmm, I'm not sure honestly, does it mean `INITIAL_PLAY_REQUESTED` can be triggered? I think it should be fine

pilievOoyala:
LGTM.

js/pulse.js
aeng7: Initial plays can be triggered any time after onAdManagerReady. This API will notify the AMC that the ad plugin is ready. Once all ad plugins are ready, the AMC will unblock the initial play, so the user can start playback by clicking on the video.

PR #206, opened 2017-08-04
Player 1340

pilievOoyala:


pilievOoyala:
I figured out another unit test I should add. Let me do that before we merge.

js/google_ima.js
```diff
@@ -330,6 +336,22 @@ require("../html5-common/js/utils/utils.js");
         }
       };
 
+      var _setNonAdRulesPreroll = privateMember(function(validAdTags)
```
aeng7: Can we include a jsdoc for this?

js/google_ima.js
```diff
@@ -330,6 +336,22 @@ require("../html5-common/js/utils/utils.js");
...
+      {
+        if (!_usingAdRules)
+        {
+          for (var index = 0; index < validAdTags.length; index++)
+          {
+            if (validAdTags[index].position == 0)
```
aeng7: Can we use === here instead of ==?

js/google_ima.js
```diff
@@ -330,6 +336,22 @@ require("../html5-common/js/utils/utils.js");
         }
       };
 
+      var _setNonAdRulesPreroll = privateMember(function(validAdTags)
+      {
+        if (!_usingAdRules)
```
aeng7: Can we add a null check for validAdTags?

test/unit-tests/ima_test.js
```diff
@@ -1839,4 +1839,20 @@ describe('ad_manager_ima', function()
       expect(_.isEmpty(sdkAdEventParams)).to.be(false);
     }
   });
+
+  it('IMA ad preloading: testing page level param', function()
```
aeng7: Do we have unit tests for the preloading (early ad request) functionality already? If not, can we add some?

js/google_ima.js
```diff
@@ -536,10 +559,14 @@ require("../html5-common/js/utils/utils.js");
           //if we are trying to play an linear ad then we need to request the ad now.
           if (this.currentAMCAdPod.ad.forced_ad_type != _amc.ADTYPE.NONLINEAR_OVERLAY)
           {
-            //reset adRequested and adTagUrl so we can request another ad
-            _resetAdsState();
-            this.adTagUrl = this.currentAMCAdPod.ad.tag_url;
-            _trySetupAdsRequest();
+            //if this is the preroll we preloaded then don't try to request it again.
+            if (!(this.preloadAds && this.numAdsPlayed === 1 && this.currentAMCAdPod.position === 0))
```
aeng7: Can you explain this condition to me?

pilievOoyala: I can't sneak anything past you :).  With ad preloading, we only preloads the first preroll if it exists.  So this is checking if the first ad to play is a preroll and we are preloading, then we have already requested the ad, so skip requesting it again.

Now that I've explained it, I probably need to go back and figure out a more generic way to mark ads in the AMC as being preloaded and then we just check a boolean.  Then we can reuse this for anything.

PR #210, opened 2017-08-17
[player-1374a]

aeng7:
IMA
-setting a default loadVideoTimeout of 15 seconds
-removed a redundant unit test

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -52,6 +52,7 @@ require("../html5-common/js/utils/utils.js");
       //Constants
       var DEFAULT_IMA_IFRAME_Z_INDEX = 10004;
       var DEFAULT_ADS_REQUEST_TIME_OUT = 15000;
+      var DEFAULT_LOAD_VIDEO_TIME_OUT = 15000;
       var AD_RULES_POSITION_TYPE = 'r';
       var NON_AD_RULES_POSITION_TYPE = 't';
       var NON_AD_RULES_PERCENT_POSITION_TYPE = 'p';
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -52,6 +52,7 @@ require("../html5-common/js/utils/utils.js");
       //Constants
       var DEFAULT_IMA_IFRAME_Z_INDEX = 10004;
       var DEFAULT_ADS_REQUEST_TIME_OUT = 15000;
+      var DEFAULT_LOAD_VIDEO_TIME_OUT = 15000;
```
pilievOoyala: What about the ad load timeout in the AMC.  IMA's load timeout and request timeout together are longer.  Won't these conflict?

js/google_ima.js
```diff
@@ -1278,6 +1279,7 @@ require("../html5-common/js/utils/utils.js");
         }
         // https://developers.google.com/interactive-media-ads/docs/sdks/googlehtml5_apis_v3#ima.AdsRenderingSettings
         var adsSettings = new google.ima.AdsRenderingSettings();
+        adsSettings.loadVideoTimeout = DEFAULT_LOAD_VIDEO_TIME_OUT;
```
pilievOoyala: Does this take care of my question above?

js/google_ima.js
aeng7: We're just matching an IMA SDK setting to what JW Player sets as default. This is to prevent 402 errors, which FloSports saw a lot of. Doc here:
https://developers.google.com/interactive-media-ads/docs/sdks/html5/v3/apis#ima.AdsRenderingSettings.loadVideoTimeout

Error 402 doc here:
https://support.google.com/dfp_premium/answer/4442429?hl=en

aeng7: There are a lot of timeouts in play and its hard to sync them all up. The 15 second ad request timeout is something we handle in the IMA plugin. The IMA SDK has a vastLoadTimeout for each wrapper:

https://developers.google.com/interactive-media-ads/docs/sdks/html5/v3/apis#ima.AdsRequest.vastLoadTimeout

and theres the loadVideoTimeout. We can revisit syncing all of these later.

PR #211, opened 2017-08-29
Player 1380

dallen42:


js/google_ima.js
```diff
@@ -1572,6 +1641,17 @@ require("../html5-common/js/utils/utils.js");
           case eventType.USER_CLOSE:
           case eventType.SKIPPED:
           case eventType.COMPLETE:
+            if(_usingAdRules)
+            {
+              this.adResponseTime = new Date().valueOf();
+            }
+            var adSkipped = false;
+            if (adEvent.type == eventType.SKIPPED)
```
pilievOoyala: We try to do === on everything unless it specifically needs a ==

js/google_ima.js
```diff
@@ -1623,6 +1703,40 @@ require("../html5-common/js/utils/utils.js");
             }
 
             break;
+          case eventType.IMPRESSION:
+
+            this.currentImpressionTime = new Date().valueOf();
+            var loadTime = this.currentImpressionTime - this.adResponseTime;
+            var protocol = "VAST";
+            if (ad && ad.g && ad.g.vpaid == true)
```
pilievOoyala: Need a ===

js/google_ima.js
dallen42: fixed

js/google_ima.js
dallen42: fixed

js/google_ima.js
```diff
@@ -751,6 +760,7 @@ require("../html5-common/js/utils/utils.js");
        */
       var _onInitialPlayRequested = privateMember(function()
       {
+        this.initialPlayRequestTime = new Date().valueOf();
```
aeng7: Although valueOf returns the same thing as getTime, it looks like valueOf is usually reserved for Javascript internals. If there are other changes to be made to this file, can we also change calls to valueOf to getTime?

js/google_ima.js
```diff
@@ -1623,6 +1703,40 @@ require("../html5-common/js/utils/utils.js");
             }
 
             break;
+          case eventType.IMPRESSION:
+
+            this.currentImpressionTime = new Date().valueOf();
+            var loadTime = this.currentImpressionTime - this.adResponseTime;
```
aeng7: Did we want to track from ad request or ad response? Just double checking.

js/google_ima.js
```diff
@@ -1623,6 +1703,40 @@ require("../html5-common/js/utils/utils.js");
...
+            var protocol = "VAST";
+            if (ad && ad.g && ad.g.vpaid === true)
+            {
+              protocol = "VPAID";
+            }
+            var type = "unknown";
+            if (ad && ad.isLinear())
+            {
+              if (ad.getContentType().lastIndexOf("video", 0 ) === 0 )
```
aeng7: Does this check check all instances of video content type?

test/unit-test-helpers/mock_ima.js
```diff
@@ -15,6 +15,8 @@ google =
     },
     Ad : function()
     {   //see https://developers.google.com/interactive-media-ads/docs/sdks/html5/v3/apis#ima.Ad
+
+      this.g = { vpaid : false };
```
aeng7: What is this.g?

dallen42: It's a container in the Ad object returned by the sdk... Google does not provide a getter function for the vpaid property stored there for some reason, so I had to grab it from the object directly. See line 1711 in google_ima.js. I realize it is hackish, but it seemed the best way to tell if the ad was vpaid or not

js/google_ima.js
dallen42: That is the idea. I couldn't find a list of all possible values getContentType() returns. Most values seem to be of the format : <media type>/<file format>.  Where <media type> is a general class like video or image, and file format is mp3, png, etc...

test/unit-test-helpers/mock_ima.js
aeng7: That's fine, works for me!

js/google_ima.js
aeng7: Sounds good.

js/google_ima.js
dallen42: The Analytics Events MVP doc defines load time to be: time between Ad Request Success and STARTED/first frame. I took Ad Request Success to mean when that the ad response is retrieved without error, which is what this.adResponseTime stores.

PR #212, opened 2017-08-30
Player 1375

aeng7:
IMA plugin
-repurposed page level setting “vpaidMode” to allow the user to specify 'insecure' or 'disabled.' Default vpaidMode is 'enabled'
-Moved setting google.ima.settings to loadMetadata so that we can read the metadata/page level settings for the google.ima.settings


aeng7:
Closing this right now as it is not needed for 4.17.x

PR #215, opened 2017-09-11
[player-1545] Adding in variable to enable skippable ads for iOS10+

pilievOoyala:
Need to verify that iPad is still okay and that all events for exitingFullscreen get propogated to the rest of the player.  There also seems to be a race condition for trying to exit fullscreen but we are going to live with it for the moment as it's only for midroll, so we are going to advice customers only use prerolls if they want skippable ads.


aeng7:
@pilievOoyala will be fixing and adding unit tests

test/unit-test-helpers/mock_ima.js
```diff
@@ -12,6 +12,7 @@ google =
       google.ima.linearAds = true;
       google.ima.delayAdRequest = false;
       google.ima.adsRenderingSettingsInstance = null;
+      google.ima.disableCustomPlaybackForIOS10Plus = false;
```
aeng7: Can you defined this up above as well? This way it does not start out as undefined.

js/google_ima.js
```diff
@@ -630,7 +637,10 @@ require("../html5-common/js/utils/utils.js");
           {
             //resumeAd will only be called if we have exited fullscreen
             //so this is safe to call
-            this.sharedVideoElement.webkitEnterFullscreen();
+            if (OO.isIosMajorVersion < 10)
```
aeng7: Can we use a separate function to calculate if we support inline playback on iOS? That way we can use this new function here and down below when we exit full screen.

js/google_ima.js
```diff
@@ -1535,6 +1550,12 @@ require("../html5-common/js/utils/utils.js");
               }
               //Since IMA handles its own UI, we want the video player to hide its UI elements
               _amc.hidePlayerUi(this.showAdControls, false);
+
+              //in the case where skippable ads are enabled we want to exit fullscreen
+              //because custom playback is disabled and ads can't be rendered in fullscreen.
+              if (OO.iosMajorVersion >= 10 && this.enableIosSkippableAds === true) {
+                this.sharedVideoElement.webkitExitFullscreen();
```
aeng7: See above comment for webkitEnterFullscreen about using a separate function.

aeng7:
LGTM

aeng7:
Single review merging to get it in this build

PR #216, opened 2017-09-21
[pbw-6815]

aeng7:
-removed an older workaround where we end the ad pod for live streams when we receive the COMPLETE event from IMA


aldigun:
Also check, if I made inappropriate comment and let me know to update it.

PR #218, opened 2017-10-05
[player-1788]

aeng7:
-muted autoplay functionality for IMA (Vast is handled via VTC)

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1601,6 +1601,7 @@ require("../html5-common/js/utils/utils.js");
 
             if (ad.isLinear())
             {
+              _linearAdIsPlaying = true;
               _amc.focusAdVideo();
             }
             break;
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -2387,6 +2389,35 @@ require("../html5-common/js/utils/utils.js");
     };
 
     /**
+     * Checks to see if autoplay requires the video to be muted
+     * @public
+     * @method TemplateVideoWrapper#requiresMutedAutoplay
+     * @param {boolean} true if video must be muted to autoplay, false otherwise
+     */
+    this.requiresMutedAutoplay = function() {
+      return OO.isSafari || OO.isIos || OO.isAndroid;
```
jpfranco: Do we want to target only Safari 11 here, or just be consistent across versions?

aeng7: I'll add that in.

aldigun:
LGTM

PR #220, opened 2017-10-20
[player-1264] Changing VAST and IMA to leave calculating ad positions…

pilievOoyala:
…, that are percentages, up to the AMC.  They will just pass along the position type and the duration of the content.

This needs to be merged at the same time as https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/307/overview

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -833,7 +833,7 @@ OO.Ads.manager(function(_, $) {
       var ad = amcAd.ad;
       this.amc.notifyPodStarted(amcAd.id, 1);
 
-      this.currentAdBeingLoaded = ad;
+      this.currentAdBeingLoaded = amcAd;
       this.loadUrl(ad.tag_url);
       loadedAds = true;
       return loadedAds;
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -909,21 +909,25 @@ OO.Ads.manager(function(_, $) {
             adManager: this.name,
             ad: adMetadata,
             duration: 0,
-            adType: this.amc.ADTYPE.AD_REQUEST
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -416,13 +416,15 @@ require("../html5-common/js/utils/utils.js");
                   "adManager": this.name,
                   "ad": ad,
                   "streams": streams,
-                  "adType": _amc.ADTYPE.UNKNOWN_AD_REQUEST
+                  "adType": _amc.ADTYPE.UNKNOWN_AD_REQUEST,
+                  "mainContentDuration": this.mainContentDuration
                 };
 
                 //percentage position types require a different calculation.
```
jpfranco: Would probably just update this comment to say that this is handled by the AMC, looks good otherwise.

pilievOoyala: Removed. since it no longer applies.

jpfranco:
LGTM

js/ad_manager_vast.js
```diff
@@ -909,21 +909,25 @@ OO.Ads.manager(function(_, $) {
...
               adData.position = +adMetadata.time / 1000;
             } else if (_isValidPosition(adMetadata.position)) {
               adData.position = +adMetadata.position / 1000;
             }
-          } else if (adMetadata.position_type == 'p') {
+          } else if (adMetadata.position_type === 'p') {
             if (_isValidPosition(adMetadata.position)) {
-              adData.position = +adMetadata.position / 100 * this.mainContentDuration;
+              adData.positionType = adMetadata.position_type;
+              adData.position = +adMetadata.position;
```
aldigun: what is = + operation?

PR #221, opened 2017-10-23
[PBW-6889] Setting VIDEO_STATE_PLAYING when content is resumed

jpfranco:
_VIDEO_STATE_PLAYING_ needs to be set when the video resumes after being paused. I added the missing handler for this.


pilievOoyala:
LGTM

PR #222, opened 2017-10-25
Player 1788b

aeng7:
-workaround of an IMA issue where the ad starts paused on Safari 11 with our muted autoplay flow
-modified the setVolume function to set the volume as long as the IMA Ads Manager exists, and will toss the previous saved volume when doing so. This allows us to use this function to prime future ads for unmuted autoplay when called on a user click

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -668,8 +668,11 @@ require("../html5-common/js/utils/utils.js");
 
       this.setVolume = function(volume)
       {
-        if (_IMAAdsManager && _linearAdIsPlaying)
+        if (_IMAAdsManager)
         {
+          if (typeof this.savedVolume === "number") {
+            this.savedVolume = -1;
+          }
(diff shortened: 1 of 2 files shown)


pilievOoyala:
LGTM

js/google_ima.js
```diff
@@ -668,8 +668,11 @@ require("../html5-common/js/utils/utils.js");
 
       this.setVolume = function(volume)
       {
-        if (_IMAAdsManager && _linearAdIsPlaying)
+        if (_IMAAdsManager)
         {
+          if (typeof this.savedVolume === "number") {
```
jpfranco: Should this be `!==`?

aeng7: Nope, because I'm getting rid of the savedVolume at this point. I technically don't need this type check though. I'll remove it.

jpfranco: Oh ok I see

PR #224, opened 2017-10-28
[player-1788]

aeng7:
-workarounds of two instances where IMA does not fire volume events
-fixed an issue where ads would start paused on mobile because IMA was not resuming when calling pause then resume immediately after
-making use of the new muteState argument since IMA does not separate mute states and volume
-adjusted the logic for saved volumes when muting/unmuting

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -674,6 +674,12 @@ require("../html5-common/js/utils/utils.js");
         {
           this.savedVolume = -1;
           _IMAAdsManager.setVolume(volume);
+          //workaround of an IMA issue where we don't receive a VOLUME_CHANGED ad event
+          //on when sharing video elements, so we'll notify of current volume and mute state now
+          if (this.videoControllerWrapper && this.sharedVideoElement)
+          {
+            this.videoControllerWrapper.raiseVolumeEvent();
+          }
         }
         else
         {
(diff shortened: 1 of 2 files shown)


js/google_ima.js
```diff
@@ -2426,7 +2438,10 @@ require("../html5-common/js/utils/utils.js");
     this.mute = function() {
       //Note there is no "mute" API. IMA seems to handle the muted
       //attribute themselves when volume is set to 0
-      volumeWhenMuted = _ima.getVolume();
+      var currentVolume = _ima.getVolume();
+      if (currentVolume) {
+        volumeWhenMuted = currentVolume;
```
jpfranco: Will this fail to set `volumeWhenMuted` when volume is 0?

js/google_ima.js
```diff
@@ -2436,7 +2451,7 @@ require("../html5-common/js/utils/utils.js");
      * @method TemplateVideoWrapper#unmute
      */
     this.unmute = function() {
-      _ima.setVolume(volumeWhenMuted);
+      _ima.setVolume(volumeWhenMuted ? volumeWhenMuted : 1);
```
jpfranco: Do we need to handle the case where the video is muted while volume is 0?

jpfranco: Ok nevermind, looking at the unit tests below it looks like this is deliberate right?

aeng7: Yep, this is deliberate. IMA considers 0 as muted.

js/google_ima.js
aeng7: Yea, but this is deliberate again. We want to store the volume when we mute, but if the current volume is alerady 0, we don't want to store it. When we restore the volume, it will be whatever value we grab here, or if we didn't grab the value, we set it to 1 (see the other line you commented on)

js/google_ima.js
pilievOoyala: This covers the case where you mute the player twice without unmuting in between, right?

pilievOoyala:
LGTM

PR #226, opened 2017-11-03
[player-2136]

aeng7:
-removed a workaround that is no longer needed for Google IMA where we had to pause and resume the ad immediately to start muted autoplayback. This was also causing an issue for podded ads on iOS


pilievOoyala:
LGTM

PR #228, opened 2017-11-07
[player-2055] Adding in a clearTimeout when we have an IMA error. And…

pilievOoyala:
… moving a different clearTimeout to earlier in the same function to not run into issues.


aeng7:
LGTM

PR #230, opened 2017-12-01
send ad tag url for ad completed event to amc

dallen42:


aeng7:
Can we add a unit test for this?

dallen42:
@aeng7 unit test added

aeng7:
LGTM

PR #231, opened 2017-12-05
[player-2281] Adding in new override for IMA that sets the max number…

pilievOoyala:
… of redirects.

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -311,6 +311,12 @@ require("../html5-common/js/utils/utils.js");
           this.enableIosSkippableAds = metadata.enableIosSkippableAds;
         }
 
+        //we don't set a default because we want Google's default if it isn't specified.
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -1117,6 +1123,10 @@ require("../html5-common/js/utils/utils.js");
 
           google.ima.settings.setDisableCustomPlaybackForIOS10Plus(this.enableIosSkippableAds);
 
+          if (this.maxRedirects && typeof (this.maxRedirects) === 'number' && this.maxRedirects > 0) {
```
jpfranco: Not sure if needed, but do we need to handle numerical string values gracefully or would it be ok to ignore them in this case?

jpfranco:
Just added a question, looks good otherwise.

aeng7:
LGTM, I'll approve but please look at JP's question. Feel free to self merge

PR #233, opened 2017-12-08
[player-2348] playerControlsOverAds now overrides the Google IMA show…

pilievOoyala:
…AdControls setting if set to true. So it will force the player controls up. There doesn't seem to be a use case where you would have playerControlsOverAds = true and showAdControls = false


aeng7:
LGTM

PR #234, opened 2017-12-09
[player-2426]

aeng7:
-we will always muted autoplay IMA starting from chrome 65 rather than 64
-fixed an issue where we attempt muted autoplay for an ad rules ad without a preroll

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -869,11 +869,12 @@ require("../html5-common/js/utils/utils.js");
        */
       var _tryStartAdsManager = privateMember(function()
       {
-        //We do not want to mute if we are using ad rules, are handling the initial ad request
+        //PLAYER-2426: We do not want to mute if we are using ad rules, are handling the initial ad request
         //for ad rules, and found no prerolls.
+        var noPrerollAdRulesAdRequest = _usingAdRules && !this.hasPreroll && this.currentAMCAdPod &&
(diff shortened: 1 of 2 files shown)


js/google_ima.js
```diff
@@ -869,13 +869,18 @@ require("../html5-common/js/utils/utils.js");
        */
       var _tryStartAdsManager = privateMember(function()
       {
-        if (!this.capturedUserClick && this.videoControllerWrapper && this.requiresMutedAutoplay())
+        //We do not want to mute if we are using ad rules, are handling the initial ad request
+        //for ad rules, and found no prerolls.
+        if (!this.capturedUserClick && this.videoControllerWrapper && this.requiresMutedAutoplay() &&
+            !(_usingAdRules && !this.hasPreroll && this.currentAMCAdPod &&
+            this.currentAMCAdPod.adType === _amc.ADTYPE.UNKNOWN_AD_REQUEST))
```
jpfranco: It might help with clarity if this condition was set to a variable (assuming there's a word that can describe this group of conditions). It's a bit hard to reason about what the `if` is checking.

pilievOoyala: werd. I agree.

pilievOoyala:
LGTM except for the one thing.

js/google_ima.js
aeng7: Sounds good. I'll make this change.

PR #237, opened 2018-01-11
[player-2621]

aeng7:
-IMA ad plugin will now report to the IMA SDK if we are intending to playback an ad muted

### DIFF
--- a/test/unit-tests/ima_test.js
+++ b/test/unit-tests/ima_test.js
@@ -2428,6 +2428,64 @@ describe('ad_manager_ima', function()
     expect(ima.requiresMutedAutoplay()).to.be(false);
   });
 
+  it('Muted Autoplay: Ad plugin notifies IMA SDK of intent to play a muted ad if muted autoplay is required', function()
+  {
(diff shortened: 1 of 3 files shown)


test/unit-tests/ima_test.js
```diff
@@ -2428,14 +2428,72 @@ describe('ad_manager_ima', function()
...
+  });
+
+  it('Muted Autoplay: Ad plugin notifies IMA SDK of intent to play a non-muted ad if muted autoplay is required but we have captured a user click', function()
+  {
+    ima.requiresMutedAutoplay = function() {
+      return true;
+    };
+    initialize(false);
+    createVideoWrapper(vci);
+    play(true);
```
jpfranco: Is this `true` indicating that we captured a user click?

dallen42:
LGTM

test/unit-tests/ima_test.js
aeng7: The true indicates that it is autoplaying. The user click is the line:

ima.setupUnmutedPlayback();

below

PR #239, opened 2018-01-23
set volumeWhenMuted during GoogleIMAVideoWrapper#setVolume

dallen42:


aeng7:
@dallen42 can we add a unit test for the use case this is solving? Thanks!

dallen42:
@aeng7 added a unit test

aeng7:
LGTM

js/google_ima.js
```diff
@@ -2598,6 +2598,7 @@ require("../html5-common/js/utils/utils.js");
      */
     this.setVolume = function(volume, muteState)
     {
+      volumeWhenMuted = volume;
```
pilievOoyala: shouldn't this be saved when we are actually about to mute and not everytime we set the volume?

aeng7: The bug was because we were doing just that. We could rename the variable to savedVolume to avoid confusion.

dallen42: ..except there is already a savedVolume in the GoogleIMA class. Might still be confusing. Will try and come up with something suitable

aeng7: Shoot. Wonder if we can consolidate them in a future release.

pilievOoyala:
LGTM

PR #240, opened 2018-01-24
[player-2676]

aeng7:
-fixed an issue where the IMA ad plugin flag for capturing a user click is reset when a content is autoplayed


pilievOoyala:
LGTM

PR #241, opened 2018-01-30
[player-2810]

aeng7:
-IMA will no longer notify of a volume change event when IMA mutes so that current player volume is not lost when IMA mutes


pilievOoyala:
LGTM

PR #245, opened 2018-03-28
[PLAYER-2914] - SSAI Web Plugin

ipadilla4:
https://jira.corp.ooyala.com/browse/PLAYER-2914

### DIFF
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -474,7 +474,7 @@ OO.Ads.manager(function(_, $)
      */
     this.onReplay = function()
     {
-        currentOffset = 0;
+      currentOffset = 0;
       this.currentAd = null;
     };
 
@@ -486,7 +486,7 @@ OO.Ads.manager(function(_, $)
      */
     var _handleId3Ad = _.bind(function(id3Object)
     {
-      // Will call _sendRequest() once live team fixes ads proxy issue.Will directly call onResponse() for now.
+      // Will call _sendRequest() once live team fixes ads proxy issue. Will directly call onResponse() for now.
       if (!this.testMode)
       {
         _sendRequest(requestUrl, id3Object);
@@ -682,7 +682,7 @@ OO.Ads.manager(function(_, $)
 
     var _onContentChanged = function()
     {
-        currentOffset = 0;
+      currentOffset = 0;
     };
 
     // Helper Functions
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -273,7 +273,7 @@ OO.Ads.manager(function(_, $)
       {
         adMode = true;
         this.currentAd = ad;
-        if (ad.ad && ad.ad.data) {
+        if (ad.ad && ad.ad.data && ad.ad.data.id) {
           this.adIdDictionary[ad.ad.data.id].curAdId = ad.id;
           _handleTrackingUrls(this.currentAd, ["impression", "start"]);
           if (ad.duration && !_.isNumber(ad.duration)) {
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -486,7 +486,7 @@ OO.Ads.manager(function(_, $)
      */
     var _handleId3Ad = _.bind(function(id3Object)
     {
-      // Will call _sendRequest() once live team fixes ads proxy issue. Will directly call onResponse() for now.
+      // Will call _sendRequest() once live team fixes ads proxy issue.Will directly call onResponse() for now.
       if (!this.testMode)
       {
         _sendRequest(requestUrl, id3Object);
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -574,7 +574,7 @@ OO.Ads.manager(function(_, $)
      * @method SsaiPulse#_notifyAmcToPlayAd
      */
     var _notifyAmcToPlayAd = _.bind(function(id3Object, adObject) {
-      if (adObject)
+      if (adObject && id3Object)
       {
         var ssaiAd = _configureSsaiObject(adObject);
         _setVastDataToDictionary(id3Object, adObject);
@@ -752,10 +752,26 @@ OO.Ads.manager(function(_, $)
       return url;
     };
 
+    /**
+    * Checks if current ID3 tag is the last one for an ad, value is 
+    * represented in percentage, being 100 the completed time.
+    * @private
+    * @method SsaiPulse#isId3ContainsCompletedTime
+    * @param  {float} id3ObjectTime  Time value from currentId3Object
+    * @returns {boolean}  True if ID3 tag time is 100
+    */
     var isId3ContainsCompletedTime = function(id3ObjectTime) {
       return id3ObjectTime === 100;
     };
     
+    /**
+    * Checks if current ID3 tag is the first one for an ad, value is
+    * represented in percentage, being 0 the start time.
+    * @private
+    * @method SsaiPulse#isId3ContainsStartedTime
+    * @param  {float} id3ObjectTime  Time value from currentId3Object
+    * @returns {boolean}  True if ID3 tag time is 0
+    */
     var isId3ContainsStartedTime = function(id3ObjectTime) {
       return id3ObjectTime === 0;
     };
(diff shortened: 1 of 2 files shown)


js/ssai_pulse.js
```diff
@@ -427,9 +474,8 @@ OO.Ads.manager(function(_, $)
      */
     this.onReplay = function()
     {
-      currentOffset = 0;
+        currentOffset = 0;
```
dallen42: Some unnecessary indentation here

js/ssai_pulse.js
```diff
@@ -440,13 +486,10 @@ OO.Ads.manager(function(_, $)
      */
     var _handleId3Ad = _.bind(function(id3Object)
     {
-      // Will call _sendRequest() once live team fixes ads proxy issue. Will directly call onResponse() for now.
+      // Will call _sendRequest() once live team fixes ads proxy issue.Will directly call onResponse() for now.
```
dallen42: Spacing issue between "issue." and "Will" was added here

js/ssai_pulse.js
```diff
@@ -593,16 +676,13 @@ OO.Ads.manager(function(_, $)
...
       this.adIdDictionary = {};
       _removeAMCListeners();
     };
 
     var _onContentChanged = function()
     {
-      currentOffset = 0;
-      this.currentAd = null;
-      this.currentId3Object = null;
+        currentOffset = 0;
```
dallen42: Remove indentation here

js/ssai_pulse.js
```diff
@@ -237,19 +273,20 @@ OO.Ads.manager(function(_, $)
...
-          amc.notifyLinearAdStarted(this.currentAd.id,
-            {
-              name: this.currentAd.ad.name,
-              hasClickUrl: true,
-              duration: this.currentAd.duration,
-              ssai: this.currentAd.ad.ssai,
-              isLive: this.currentAd.ad.isLive
-            }
-          );
+          if (ad.ad.name && ad.duration && ad.ad.ssai && ad.ad.isLive) {
```
dallen42: @cacel when I asked to validate some of these fields, I did not mean to just check for existence or if they are true. In fact, this line might break the code. What I mean by data validation is to ensure that ad.duration is a number, for example. Not all fields may need to be validated this way, only some of them. I would say duration is the main one to focus on

js/ssai_pulse.js
```diff
@@ -209,13 +230,28 @@ OO.Ads.manager(function(_, $)
...
+      if (!amc.isLiveStream) 
       {
-        currentOffset = offset;
+        if (duration && duration > 0)
+        {
+          offsetParam = duration - playhead;
+        }
+      }
+      //For live streams, if user moved the playback head into the past, offset is the seconds in the past that user is watching
+      if ((amc.isLiveStream && offset && duration) && offset > 0 && offset < duration)
```
dallen42: @cacel same thing on validation here. Existence of `offset` and `duration` is not enough, we should verify that these are actually numeric as we do arithmetic operations on them later and if they are strings or some non numeric data then this could cause an exception

js/ssai_pulse.js
```diff
@@ -209,13 +230,28 @@ OO.Ads.manager(function(_, $)
...
+      if (!amc.isLiveStream) 
       {
-        currentOffset = offset;
+        if (duration && typeof duration === 'number' && duration > 0)
+        {
+          offsetParam = duration - playhead;
+        }
+      }
+      //For live streams, if user moved the playback head into the past, offset is the seconds in the past that user is watching
+      if ((amc.isLiveStream && (offset && typeof offset === 'number') && (duration && typeof duration === 'number')) && offset > 0 && offset < duration)
```
dallen42: @cacel underscore has a `_.isNumber` function that we use to determine if a value is numeric. You can see some examples in ad_manager_vast.js. Please use that instead of `typeof`

js/ssai_pulse.js
```diff
@@ -237,19 +273,17 @@ OO.Ads.manager(function(_, $)
       {
         adMode = true;
         this.currentAd = ad;
-        if (this.currentAd.ad)
-        {
-          this.currentAd.ad.id3AdId = this.currentId3Object.adId;
+        if (ad.ad && ad.ad.data && (ad.duration && typeof ad.duration === 'number')) {
```
dallen42: @cacel I did some research on where this event goes, and I don't think we need to fail if ad.duration is not a number. The only place I found it used was in analytics as part of a metadata object. Instead of placing this validation check in the if statement (also use `_.isNumber` here for this), you should put it inside the if and assign some dummy value like 0 or -1 if the check fails

js/ssai_pulse.js
```diff
@@ -237,19 +273,17 @@ OO.Ads.manager(function(_, $)
...
+          this.adIdDictionary[ad.ad.data.id].curAdId = ad.id;
```
dallen42: @cacel @ipadilla4  we might want to also check if `ad.ad.data.id` exists

js/ssai_pulse.js
```diff
@@ -381,42 +415,53 @@ OO.Ads.manager(function(_, $)
     this.onVideoTagFound = function(eventName, videoId, tagType, metadata)
     {
       OO.log("TAG FOUND w/ args: ", arguments);
-      this.currentId3Object = _parseId3Object(metadata);
-      if (this.currentId3Object)
+      if(!amc.isLiveStream && !currentOffset)
```
dallen42: @ipadilla4 why was this block added back here? This effectively ignores ID3 tags for VOD, was there a reason to add this back to the code?

js/ssai_pulse.js
dallen42: @cacel @ipadilla4 is ad.ad.data.id guaranteed to exist here? I suspect it should, (especially if ad.ad.data exists) but we might want to add a null check just in case

js/ssai_pulse.js
```diff
@@ -237,19 +273,20 @@ OO.Ads.manager(function(_, $)
       {
         adMode = true;
         this.currentAd = ad;
-        if (this.currentAd.ad)
-        {
-          this.currentAd.ad.id3AdId = this.currentId3Object.adId;
+        if (ad.ad && ad.ad.data) {
+          this.adIdDictionary[ad.ad.data.id].curAdId = ad.id;
```
dallen42: @cacel @ipadilla4 is `ad.ad.data.id` guaranteed to exist here? I suspect it should, (especially if `ad.ad.data` exists) but we might want to add a null check just in case

ipadilla4: @dallen42 it should exist but I agree to verify just in case, I will add that change.

js/ssai_pulse.js
```diff
@@ -475,46 +561,44 @@ OO.Ads.manager(function(_, $)
...
+      return ssaiAd;
+    };
+    
+    /**
+     * Force an ad to play with configured ssai ad data
+     * @private
+     * @method SsaiPulse#_notifyAmcToPlayAd
+     */
+    var _notifyAmcToPlayAd = _.bind(function(id3Object, adObject) {
+      if (adObject)
```
aeng7: Can we add a null check for id3Object as well?

js/ssai_pulse.js
```diff
@@ -671,6 +752,14 @@ OO.Ads.manager(function(_, $)
       return url;
     };
 
+    var isId3ContainsCompletedTime = function(id3ObjectTime) {
```
aeng7: Can we add jsdocs for this function and the one below?

js/ssai_pulse.js
```diff
@@ -1029,6 +1120,11 @@ OO.Ads.manager(function(_, $)
       return currentOffset;
     }, this);
 
+    this.setCurrentOffset = function(offset)
```
aeng7: Can we add jsdocs?

js/ssai_pulse.js
```diff
@@ -1121,34 +1217,49 @@ OO.Ads.manager(function(_, $)
...
-    var _adEndedCallback = _.bind(function()
-    {
-      _clearAdDurationTimeout();
-      if (this.currentAd)
-      {
-        amc.notifyLinearAdEnded(this.currentAd.id);
-        amc.notifyPodEnded(this.currentAd.id);
-        _handleTrackingUrls(this.currentAd, ["firstQuartile", "midpoint", "thirdQuartile", "complete"]);
+    var _handleImpressionCalls = _.bind(function(curId3Object) {
+      if (!isId3ContainsStartedTime(curId3Object.time)) {
```
aeng7: For my own curiosity, does this mean that we are sending out the impression when we receive the first id3 tag after the playback starts? Does only the initial id3 tag to prompt ad playback contain the start time?

aeng7:
LGTM

PR #247, opened 2018-04-23
[PLAYER-2104] Removing unnecessary references to InitOOJQuery

jpfranco:
This fixes the issue in which the global jQuery object is either destroyed or modified by the Ad Manager Vast or SSAI Pulse plugins. The `InitOOJQuery` module from `html5-common` is not actually needed because ad plugins get their jQuery reference from the AMC. We will be deprecating `InitOOJQuery` since it causes conflicts with the global jQuery and it's only being used for unit test initialization.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -5,7 +5,6 @@
 
 
 require("../html5-common/js/utils/InitModules/InitOO.js");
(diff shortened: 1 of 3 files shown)


aeng7:
LGTM

pilievOoyala:
LGTM

PR #248, opened 2018-04-25
Player 2917 [SSAI]

dallen42:


js/ssai_pulse.js
```diff
@@ -416,14 +423,27 @@ OO.Ads.manager(function(_, $)
...
+      if (!amc.isLiveStream && !currentOffset)
       {
         return null;
       }
+
       var currentId3Object = _parseId3Object(metadata);
       if (currentId3Object)
       {
+
+        if (currentId3Object["time"] < 100) {
```
pilievOoyala: What is 100? 100 seconds, 100ms?  Can you make it a constant otherwise it kind of looks like a magic number. And you reuse it in the next if.

js/ssai_pulse.js
```diff
@@ -602,6 +630,16 @@ OO.Ads.manager(function(_, $)
...
     };
 
+    /**
+     * Called if the ajax call for SSAI metadata fails
+     * @public
+     * @method SsaiPulse#onMetadataError
+     */
+    this.onMetadataError = function(error)
+    {
+      OO.log("SSAI Metadata Request: Error" + JSON.stringify(error));
```
pilievOoyala: What do we normally send when metadata fails in the core?  Should we send the same type of message here to the core, so IQ can record it?

js/ssai_pulse.js
```diff
@@ -710,6 +748,29 @@ OO.Ads.manager(function(_, $)
...
 
+    /**
+     * Parses the ad url to obtain the ssai guid and embed code
+     * @private
+     * @method SsaiPulse#_parseUrl
+     * @param {string} url The stream url
+     */
+    var _parseUrl = _.bind(function(url)
+    {
+      var urlParts = url.split("?");
```
pilievOoyala: Can you add some null checks or a try/catch block?  Just in case someone thinks it's useful and wants to use it for something else.

dallen42: @pilievOoyala added, I think it is safe now but let me know if you see anything still missing

js/ssai_pulse.js
dallen42: @pilievOoyala right now there is no specific message we use for ad metadata failure. This technically isn't a VAST request so it is not a request error. API_ERROR might work but I would need to add the ability to send that in the AMC. Also note that none of these are reported by reporter.js currently, they are only processed by the JSON pipeline

pilievOoyala:
LGTM.

pilievOoyala:
@aeng7 @jpfranco  Please review as well.

aeng7:
@dallen42 there are conflicts that need to be resolved. Please take a look at this, thanks!

js/ssai_pulse.js
aeng7: If url is null or undefined, an error will still be thrown right? I think we need another null check for "url"

aeng7:
LGTM other than a small comment about a null check. Please look at this while fixing the merge conflicts.

js/ssai_pulse.js
dallen42: @aeng7 fixed

js/ssai_pulse.js
dallen42: @pilievOoyala added a function to let amc raise an api error

js/ssai_pulse.js
dallen42: @pilievOoyala fixed

dallen42:
@pilievOoyala @aeng7 merge conflicts are fixed, please take a look and re-approve

js/ssai_pulse.js
aeng7: If url is undefined, we'll still have an error here. We can probably check typeof url === "string" to cover all use cases so we can safely call .split on the next line.

js/ssai_pulse.js
```diff
@@ -713,6 +757,43 @@ OO.Ads.manager(function(_, $)
...
 
+    /**
+     * Parses the ad url to obtain the ssai guid and embed code
+     * @private
+     * @method SsaiPulse#_parseUrl
+     * @param {string} url The stream url
+     */
+    var _parseUrl = _.bind(function(url)
+    {
+      if (url === null) {
```
aeng7: If url is undefined we would still have an error thrown. It might be best to check if typeof url === "string" so that we can safely call .split in the next line.

aeng7:
@dallen42  added a comment.

js/ssai_pulse.js
dallen42: @aeng7 changed

PR #249, opened 2018-04-30
[Player 2678] Add player param support for SSAI

orobles-ooyala:


js/ssai_pulse.js
```diff
@@ -132,6 +132,9 @@ OO.Ads.manager(function(_, $)
     {
       amc = adManagerController;
 
+      // Request embed code provider metadata
+      amc.willRequireEmbedCodeMetadata();
```
dallen42: @orobles-ooyala can you create a unit test that checks that willRequireEmbedCodeMetadata is fired?

orobles-ooyala:
@dallen42 unit test for amc.willRequireEmbedCodeMetadata is now merged in mjolnir

dallen42:
@orobles-ooyala we also ideally want a unit test in this repo though to make sure the function will be called on the plugin side. You can see a simple example of how to do this here: https://github.com/ooyala/html5-ad-plugins/blob/master/test/unit-tests/ssai_pulse_test.js#L166-L176

Basically you need to create a mock function for `amc.willRequireEmbedCodeMetadata()` that sets some temp variable to a specific value, then call `SsaiPulse.initialize(amc)`, and then after check to make sure the variable is set to the value specified in the mock function.

pilievOoyala:
LGTM

PR #254, opened 2018-05-22
[PLAYER-3539] Rename Ooyala Player SSAI plugin

ipadilla4:
https://jira.corp.ooyala.com/browse/PLAYER-3539

### DIFF
--- a/js/ooyala_ssai.js
+++ b/js/ooyala_ssai.js
@@ -33,7 +33,7 @@ OO.Ads.manager(function(_, $)
    */
   var OoyalaSsai = function()
   {
-    this.name = "ooyala_ssai";
+    this.name = "ooyala-ssai-ads-manager";
     this.ready = false;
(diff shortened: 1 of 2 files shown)


dallen42:
@ipadilla4 we also need to update the unit test file ssai_pulse_test.js with the new names

dallen42:
@ipadilla4 also, I think it would be a good idea to create a sandbox with this code and test to make sure that the renamed plugin will load in a debug page and work correctly

ipadilla4:
Ok I will create a sandbox for this and let you know my findings.

ipadilla4:
I tested the changes and it seems to be working fine. Debug page is here: http://tiny.cc/2577ty

dallen42:
LGTM

ipadilla4:
@cacel  I searched for files using SSAI plugin and the only one referencing it is the test file, which I already included in this PR.

PR #255, opened 2018-05-29
[player-3603]

aeng7:
-fixed an issue where we are focusing the ad video element too late in environments where we’re allowed only a single video element
-workaround of an issue where the IMA iframe blocks clicks to our UI

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -1743,6 +1743,13 @@ require("../html5-common/js/utils/utils.js");
         switch (adEvent.type)
         {
           case eventType.LOADED:
+            //We normally focus the video element when receiving the ad STARTED notification from IMA.
+            //However, for environments where only a single video element is supported, we will focus the video
+            //element right before starting the ad. This allows the video element listeners to get unregistered
+            //and prevent shared video element interference.
+            if (_amc.ui.useSingleVideoElement) {
+              _amc.focusAdVideo();
+            }
             this.currentMedia = ad.getMediaUrl();
             _resetUIContainerStyle();
 
@@ -1752,6 +1759,16 @@ require("../html5-common/js/utils/utils.js");
             }
             break;
           case eventType.STARTED:
+            //Workaround of an issue on iOS where the IMA iframe is capturing clicks after an ad.
+            //We will show the iframe on receiving STARTED and hide it when receiving COMPLETE
+            if (OO.isIos) {
+              var IMAiframe = $("iframe[src^='http://imasdk.googleapis.com/']")[0];
+              if (IMAiframe && IMAiframe.style)
+              {
+                IMAiframe.style.display = 'block';
+              }
+            }
+
             if (this.videoControllerWrapper && this.requiresMutedAutoplay()) {
               //workaround of an IMA issue where we don't receive a MUTED ad event
               //on Safari mobile, so we'll notify of current volume and mute state now
@@ -1762,7 +1779,10 @@ require("../html5-common/js/utils/utils.js");
             if(ad.isLinear())
             {
               _linearAdIsPlaying = true;
-              _amc.focusAdVideo();
+
+              if (!_amc.ui.useSingleVideoElement) {
+                _amc.focusAdVideo();
+              }
 
               if (this.savedVolume >= 0)
               {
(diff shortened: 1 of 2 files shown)


js/google_ima.js
```diff
@@ -1818,6 +1838,16 @@ require("../html5-common/js/utils/utils.js");
           case eventType.USER_CLOSE:
           case eventType.SKIPPED:
           case eventType.COMPLETE:
+            //Workaround of an issue on iOS where the IMA iframe is capturing clicks after an ad.
+            //We will show the iframe on receiving STARTED and hide it when receiving COMPLETE
+            if (OO.isIos) {
+              var IMAiframe = $("iframe[src^='http://imasdk.googleapis.com/']")[0];
+              if (IMAiframe && IMAiframe.style)
+              {
+                IMAiframe.style.display = 'none';
```
aeng7: This is a slightly risky change, since we're directly modifying the style of an IMA iframe and IMA could release an update which breaks our integration with IMA. However, I could not find an alternative fix/workaround since the IMA iframe itself is capturing these clicks. Setting pointer-events none on the iframe parent does not work. The only thing that does seem to work is setting display none on the parent, but this prevents the ad from rendering under most circumstances.

aeng7: What does everyone think about this? If we don't put this workaround in, the user cannot interact with the skin outside of the control bar. This means that the giant pause or replay icons in the middle of the skin cannot be clicked if there was an IMA ad.

aeng7: Note that this PR is two bug fixes/workarounds in one. I could also split this change out into another PR if people don't agree with this change.

js/google_ima.js
```diff
@@ -1752,6 +1759,16 @@ require("../html5-common/js/utils/utils.js");
             }
             break;
           case eventType.STARTED:
+            //Workaround of an issue on iOS where the IMA iframe is capturing clicks after an ad.
+            //We will show the iframe on receiving STARTED and hide it when receiving COMPLETE
+            if (OO.isIos) {
+              var IMAiframe = $("iframe[src^='http://imasdk.googleapis.com/']")[0];
```
jpfranco: Is it possible to make this more specific by matching only iframes that are children of our player (in case there are other IMA iframes loaded on the page)?

jpfranco: I think we should also leave out the protocol if possible in order to be able to match the https url.

js/google_ima.js
jpfranco: Could we achieve something similar by tweaking the CSS of the plugin containers on the skin?

js/google_ima.js
```diff
@@ -1818,6 +1838,16 @@ require("../html5-common/js/utils/utils.js");
           case eventType.USER_CLOSE:
           case eventType.SKIPPED:
           case eventType.COMPLETE:
+            //Workaround of an issue on iOS where the IMA iframe is capturing clicks after an ad.
+            //We will show the iframe on receiving STARTED and hide it when receiving COMPLETE
+            if (OO.isIos) {
```
pilievOoyala: do we want to check isIos or do we want to check for single element video?  Otherwise this effects both iphone and ipad.  But is the issue occuring on ipad?

aeng7: I haven't tested iPad. This change is unrelated to single video element mode.

js/google_ima.js
aeng7: That would be ideal, but as mentioned above, I cannot get anything to work unless I set display:none on the plugin containers, which will mess up IMA setup. One thing I could do is have the plugin container start with display:block for its initial hidden state, and after IMA is setup, switch between display:none and display:block depending on if the plugin container should be visible. This would increase complexity of the skin. But there is still no guarantee that IMA will be ok with this since they require the container to have a calculated size (with regards to the DOM) prior to setting up for an ad. What do you think?

js/google_ima.js
aeng7: This issue is observed on iPad as well.

js/google_ima.js
jpfranco: Do we have a reference to an element from our player? If we do then we could use `element.querySelector()` without having to use the class

js/google_ima.js
pilievOoyala: ok, you can disregard this then.

pilievOoyala:
LGTM aside from JP's comments.

js/google_ima.js
jpfranco: @aeng7 hmm maybe that approach could also work with a negative z-index instead of `display: none`? I'm not sure which approach I'd prefer though, it does sound a bit complicated.

aeng7: I've tried the negative z-index. It does not work. The iframe still captures clicks. It looks to be an iOS quirk because this is observed only on iOS

js/google_ima.js
aeng7: Ah, yep, we do have a reference to the container. Let me make this change.

aeng7: Added

PR #256, opened 2018-05-30
[player-3601]

aeng7:
-fixed an issue where we are focusing the ad video element too late with Freewheel, causing video element sync issues between the video plugin and the Freewheel SDK

### DIFF
--- a/js/freewheel.js
+++ b/js/freewheel.js
@@ -519,6 +519,7 @@ OO.Ads.manager(function(_, $) {
             _registerDisplayForLinearAd();
             fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_CLICK_DETECTION, false, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
             slotStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(ad) {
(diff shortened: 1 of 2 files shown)


js/freewheel.js
```diff
@@ -519,6 +519,7 @@ OO.Ads.manager(function(_, $) {
             _registerDisplayForLinearAd();
             fwContext.setParameter(tv.freewheel.SDK.PARAMETER_RENDERER_VIDEO_CLICK_DETECTION, false, tv.freewheel.SDK.PARAMETER_LEVEL_GLOBAL);
             slotStartedCallbacks[ad.ad.getCustomId()] = _.bind(function(ad) {
+              amc.focusAdVideo();
```
jpfranco: Do we need to remove the previous call to `amc.focusAdVideo()` or was it not there before?

aeng7: It was not there before. It was done by the AMC after receiving notifyLinearAdStarted. There will be two calls to focusAdVideo, but focusing an already focused video element is fine. IMA already does this.

pilievOoyala:
LGTM. But lets wait for JP's approval

js/freewheel.js
jpfranco: Sounds good

PR #258, opened 2018-06-04
quick fix to allow use of the content url domain as the domain for th…

dallen42:
…e v1/metadata call to allow calling ssai staging api for metadata

### DIFF
--- a/js/ssai_pulse.js
+++ b/js/ssai_pulse.js
@@ -40,6 +40,7 @@ OO.Ads.manager(function(_, $)
     this.testMode = false;
     this.timeLine = {};
     this.currentEmbed = "";
+    this.domainName = "ssai.ooyala.com";
     this.ssaiGuid = "";
     
     this.currentAd = null;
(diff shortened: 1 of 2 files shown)


js/ssai_pulse.js
```diff
@@ -776,6 +777,7 @@ OO.Ads.manager(function(_, $)
       var mainUrl = urlParts[0];
       var mainUrlParts = mainUrl.split("/");
       if (mainUrlParts !== null) {
+      	this.domainName = mainUrlParts[2];
```
aeng7: Can we update the jsdoc to include that this function also parses for the domain name?

aeng7:
LGTM other than a minor comment about jsdocs

js/ssai_pulse.js
dallen42: @aeng7 added

aeng7:
LGTM

pilievOoyala:
LGTM

PR #261, opened 2018-06-08
[PLAYER-3418] Expected UI Controls during SSAI Ad Playback

ipadilla4:
Implementing ad pause/resume methods to clear ad timeout and make ui ad controls to disappear in the right time.

### DIFF
--- a/js/ooyala_ssai.js
+++ b/js/ooyala_ssai.js
@@ -327,7 +327,9 @@ OO.Ads.manager(function(_, $)
     this.pauseAd = function(ad)
     {
       //Removing the ad timeout since ad was paused
-      clearTimeout(this.adIdDictionary[ad.ad.data.id].adTimer);
+      if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id]) {
+        clearTimeout(this.adIdDictionary[ad.ad.data.id].adTimer);
+      }
     };
 
     /**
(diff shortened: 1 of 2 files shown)


aeng7:
@ipadilla4 are there any unit tests we can add?

js/ooyala_ssai.js
```diff
@@ -326,6 +326,8 @@ OO.Ads.manager(function(_, $)
      */
     this.pauseAd = function(ad)
     {
+      //Removing the ad timeout since ad was paused
+      clearTimeout(this.adIdDictionary[ad.ad.data.id].adTimer);
```
aeng7: Can we add null checks for the following?
ad
ad.ad
ad.ad.data
this.adIdDictionary[ad.ad.data.id]?

js/ooyala_ssai.js
```diff
@@ -337,6 +339,11 @@ OO.Ads.manager(function(_, $)
      */
     this.resumeAd = function(ad)
     {
+      //Setting the ad callback again since ad was resumed
+      this.adIdDictionary[ad.ad.data.id].adTimer = _.delay(
```
aeng7: Can we add null checks for the following?
ad
ad.ad
ad.ad.data
this.adIdDictionary[ad.ad.data.id]
currentId3Object
?

dallen42:
@ipadilla4 LGTM apart from what Alex already commented, will merge when updated

js/ooyala_ssai.js
ipadilla4: Done

test/unit-tests/ooyala_ssai_test.js
```diff
@@ -853,4 +853,48 @@ describe('ad_manager_ooyala_ssai', function()
...
+        }
+      }
+    };
+    OoyalaSsai.initialize(amc);
+    var currentId3Object = OoyalaSsai.onVideoTagFound("eventName", "videoId", "tagType", mockId3Tag);
+    expect(OoyalaSsai.adIdDictionary[currentId3Object.adTimer]).to.not.be(null);
+    OoyalaSsai.pauseAd(ad);
+    expect(OoyalaSsai.adIdDictionary[currentId3Object.adTimer]).to.be(undefined);
+    OoyalaSsai.resumeAd(ad);
+    expect(OoyalaSsai.adIdDictionary[currentId3Object.adTimer]).to.not.be(null);
```
aeng7: Can we check for something else other than not null? The reason for this is because undefined is not null, so the adTimer could remain undefined and this test would still pass.

aeng7:
LGTM

PR #264, opened 2018-06-20
Player 3777

dallen42:


js/ooyala_ssai.js
```diff
@@ -342,12 +347,15 @@ OO.Ads.manager(function(_, $)
      */
     this.resumeAd = function(ad)
     {
-      if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id] && currentId3Object) {
```
aeng7: Just double checking, do we no longer care about the existence of currentId3Object in this function?

aeng7:
LGTM

test/unit-tests/ooyala_ssai_test.js
```diff
@@ -292,17 +293,39 @@ describe('ad_manager_ooyala_ssai', function()
...
+    OoyalaSsai.onMuteStateChanged("muteStateChanged", true);
+    expect(trackingUrlsPinged.muteUrl).to.be(1);
+    OoyalaSsai.onMuteStateChanged("muteStateChanged", false);
+    expect(trackingUrlsPinged.unmuteUrl).to.be(1);
+
+    // volume change events should also trigger mute / unmute urls
+    OoyalaSsai.onAdVolumeChanged("adVolumeChanged", 0);
+    expect(trackingUrlsPinged.muteUrl).to.be(2);
+    OoyalaSsai.onAdVolumeChanged("adVolumeChanged", 50);
+    expect(trackingUrlsPinged.unmuteUrl).to.be(2);
```
aeng7: Will a volume change that unmutes the player cause the unmuteUrl to be pinged twice? Can this happen with our player? We probably don't want the mute/unmuteUrls to be pinged twice on one mute state change.

js/ooyala_ssai.js
dallen42: @aeng7 this was actually a bug from before I think... there is no `currentId3Object` variable in scope here. The goal of this is to get the duration of the ad to restart the ad timer with that duration. The `ad` object itself should contain this at this point so we do not actually need it even

test/unit-tests/ooyala_ssai_test.js
dallen42: @aeng7 there is logic in the plugin that keeps track of the mute state and only pings the url if a change of state is found. This is in both `onMuteStateChanged` and `onAdVolumeChanged`, although there looks like a possible but in `onAdVolumeChanged` that does not check the `isMuted` bool on setting the volume to 0 so I will change that

dallen42: @aeng7 made some changes to the logic to ensure `lastVolume` and `isMuted` are consistent and added some unit tests to check and make sure the urls aren't pinged multiple times incorrectly

js/ooyala_ssai.js
```diff
@@ -328,9 +330,12 @@ OO.Ads.manager(function(_, $)
...
       //Removing the ad timeout since ad was paused
-      if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id]) {
-        clearTimeout(this.adIdDictionary[ad.ad.data.id].adTimer);
-      }
+      if (adMode) {
+        _handleTrackingUrls(this.currentAd, ["pause"]);
+        if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id]) {
+          clearTimeout(this.adIdDictionary[ad.ad.data.id].adTimer);
+        }
+  	  }
```
jpfranco: Can we fix this indentation?

js/ooyala_ssai.js
```diff
@@ -342,12 +347,15 @@ OO.Ads.manager(function(_, $)
...
     {
-      if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id] && currentId3Object) {
-        //Setting the ad callback again since ad was resumed
-        this.adIdDictionary[ad.ad.data.id].adTimer = _.delay(
-          _adEndedCallback(null, ad.ad.data.id),
-          currentId3Object.duration * 1000
-        );
+      if (adMode) {
+        _handleTrackingUrls(this.currentAd, ["resume"]);
+        if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id] && _.isNumber(ad.duration)) {
```
jpfranco: I think `_.isNumber(ad.duration)` might return true for `NaN`. We could add a a check for `_.isFinite()` if that's the case.

js/ooyala_ssai.js
```diff
@@ -694,21 +702,53 @@ OO.Ads.manager(function(_, $)
...
-          _handleTrackingUrls(this.currentAd, ["unmute"]);
+          url = ["unmute"];
         }
-      }
+
+        if (adMode)
+        {
+          _handleTrackingUrls(this.currentAd, url);
+        }
+  	  }
```
jpfranco: Indentation issue

js/ooyala_ssai.js
dallen42: @jpfranco there were mixed spaces and tabs, fixed now

js/ooyala_ssai.js
dallen42: @jpfranco changed

js/ooyala_ssai.js
dallen42: @jpfranco fixed

aeng7:
LGTM

PR #265, opened 2018-07-03
[pbw-7283] Fixing pulse plugin not respecting forced muted autoplay f…

pilievOoyala:
…or ads.

There are no new unit tests because we still need to get the unit tests from pulse and I don't want to build a whole new file without seeing what they have created already.


aeng7:
LGTM

zelibobla:
No activity in the ticket. No pushes from support for 6 months. Consider as outdated and not required. Closing.

PR #266, opened 2018-07-09
Player 3519

pilievOoyala:
There is no unit test because this section of code is only run when we are setting up the Google SDK and we don't download that for unit tests.


js/google_ima.js
```diff
@@ -1206,7 +1206,11 @@ require("../html5-common/js/utils/utils.js");
             //These are required by Google for tracking purposes.
             google.ima.settings.setPlayerVersion(PLUGIN_VERSION);
             google.ima.settings.setPlayerType(PLAYER_TYPE);
-            google.ima.settings.setLocale(OO.getLocale());
+            if (_amc.uiLanguage) {
```
aeng7: I think for the IMA plugin, we have curly braces on new lines.

aeng7:
LGTM

PR #269, opened 2018-07-20
Player 3986 - Update Node to 8.11.3

aeng7:
Updating Node to 8.11.3 and allow for usage of ES6. DO NOT MERGE YET! Opening just for review. We'll need to update html5-common and the build scripts to accommodate for these Node updates


Makefile
```diff
@@ -6,13 +6,15 @@ UGLIFY_BIN = ./node_modules/.bin/uglifyjs
 
 test:
 	@NODE_ENV=test $(MOCHA_BIN) \
+	  --exit \
```
aeng7: Mocha no longer exits by itself if there are things to clean up. Adding this exit option will mimic the older Mocha version we were using

js/ad_manager_vast.js
```diff
@@ -2123,7 +2123,7 @@ OO.Ads.manager(function(_, $) {
       var mediaFile = linearXml.find("MediaFile");
 
       parseTrackingEvents(result.tracking, linearXml);
-      if (mediaFile.size() > 0) {
+      if (mediaFile.length > 0) {
```
aeng7: The updated jquery no longer supports the size() API. They suggest to use .length instead.

js/ad_manager_vast.js
```diff
@@ -3448,9 +3448,12 @@ OO.Ads.manager(function(_, $) {
      */
     var _updateCreativeSize = _.bind(function() {
       if (this._slot) {
+        var clientRect = this._slot.getBoundingClientRect();
```
aeng7: jsdom does not support editing offsetWidth/Height, so making use of getBoundingClientRect instead for tests.

aeng7:
The html5-common submodule is currently synced to the player-3896 branch in the html5-common repo. There is a package.json update needed in the html5-common repo that would break things unless we're using node 8.11.3. When all the Node 8.11.3 PRs are ready and approved, I'll create a PR in the html5-common repo to push the package.json change and update these other PRs with the submodule update.

js/ad_manager_vast.js
jpfranco: Wouldn't this be using the jquery version from the core at runtime, which would still be the old version?

package.json
```diff
@@ -13,27 +13,31 @@
...
-    "mime": "^1.3.4"
+    "mime": "^2.3.1"
   },
   "devDependencies": {
-    "browserify": "11.0.1",
+    "browserify": "16.2.2",
+    "babelify": "^9.0.0",
+    "@babel/core": "^7.0.0-beta.51",
+    "@babel/preset-env": "^7.0.0-beta.51",
+    "babel-core": "^7.0.0-bridge.0",
```
jpfranco: Should we use the latest stable version of Babel instead of a beta version?

js/ad_manager_vast.js
aeng7: This should still be fine. The .length returns the same value as .size(). Both of these APIs have existed since version 1.0 of jquery:

https://api.jquery.com/length/

package.json
aeng7: Babelify implies/recommends installing the latest versions (which are the beta versions) in their readme:

https://github.com/babel/babelify

babel-core might not be needed though since we are using @ babel/core. Let me check this for both the skin and this repo.

jpfranco: Oh ok, that's weird

package.json
```diff
@@ -13,27 +13,30 @@
   },
   "dependencies": {
     "aws-sdk": "^2.1.49",
-    "mime": "^1.3.4"
+    "mime": "^2.3.1"
```
jpfranco: Forgot about this when I did the initial review, but it might be worth adding the `engines` property with the minimum required node version

aeng7: Will do

aeng7: Added

js/ad_manager_vast.js
```diff
@@ -1543,7 +1543,7 @@ OO.Ads.manager(function(_, $) {
             this.amc.notifyNonlinearAdEnded(ad.id);
           }
           if((ad.ad.adPodIndex === ad.ad.adPodLength && !failed) || !nextAd) {
-            var adPod = adPodPrimary || ad.id;
+            var adPod = adPodPrimary || ad;
```
pilievOoyala: nice clean up!

pilievOoyala:
LGTM.

PR #272, opened 2018-08-02
[PLAYER-3912] Notifying non-linear ad end when stream ends with active overlay

jpfranco:
The cleanup of the non-linear ad wasn't being performed by the skin when the video was switched while the overlay was active.

### DIFF
--- a/js/ad_manager_vast.js
+++ b/js/ad_manager_vast.js
@@ -1494,6 +1494,12 @@ OO.Ads.manager(function(_, $) {
             _skipAd(currentAd);
           }
         }
+        // [PLAYER-3912]
+        // Make sure that NONLINEAR_AD_PLAYED gets fired when a stream ends
+        // with an active overlay, otherwise the skin will keep it for new videos.
+        else if (!ad.isLinear && params.code === this.amc.AD_CANCEL_CODE.STREAM_ENDED) {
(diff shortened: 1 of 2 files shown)


js/ad_manager_vast.js
```diff
@@ -1494,6 +1494,12 @@ OO.Ads.manager(function(_, $) {
             _skipAd(currentAd);
           }
         }
+        // [PLAYER-3912]
+        // Make sure that NONLINEAR_AD_PLAYED gets fired when a stream ends
+        // with an active overlay, otherwise the skin will keep it for new videos.
+        else if (!ad.isLinear && params.code === this.amc.AD_CANCEL_CODE.STREAM_ENDED) {
+          _endAd(ad, false);
```
jpfranco: Not sure whether I should use `_skipAd` or `_endAd` here, but I think `_endAd` makes more sense since skipping is usually a manual action and this is sort of a side-effect of switching videos.

aeng7:
LGTM

js/ad_manager_vast.js
pilievOoyala: I think endAd is the better thing to use. Just in case we fire off something on the message bus we don't want to report another skip for analytics.

pilievOoyala:
LGTM

PR #274, opened 2018-09-05
Fix/player 3832

zelibobla:
Added gulp webserver for debugging reasons.
Core fix: ignored 1009 'The VAST response document is empty' Google IMA SDK error returned when request for ads is cancelled.

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -17,6 +17,7 @@ require("../html5-common/js/utils/utils.js");
 
   OO.Ads.manager(function(_, $)
   {
+
     /**
      * @class GoogleIMA
      * @classDesc The GoogleIMA Ads Manager class is registered as an ads manager with the Ad Manager Controller.
@@ -531,6 +532,7 @@ require("../html5-common/js/utils/utils.js");
        */
       this.playAd = function(amcAdPod)
       {
+        
         if(this.currentAMCAdPod)
         {
           _endCurrentAd(true);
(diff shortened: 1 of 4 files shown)


js/google_ima.js
```diff
@@ -1385,6 +1385,17 @@ require("../html5-common/js/utils/utils.js");
        */
       var _onImaAdError = privateMember(function(adError)
       {
+        if (adError &&
+          adError.type === 'adError' &&
+          adError.h &&
```
pilievOoyala: Is this safe? It looks like you are accessing a the variable from the minified code which could change with their next release. There should be a function you can call to extract the ad error code.

pilievOoyala: https://developers.google.com/interactive-media-ads/docs/sdks/html5/v3/apis#ima.AdError

just do aderror.getErrorCode()

zelibobla: Thank you, Peter. Fixed this.

js/google_ima.js
```diff
@@ -1385,6 +1385,16 @@ require("../html5-common/js/utils/utils.js");
...
       var _onImaAdError = privateMember(function(adError)
       {
+        if (adError &&
+          adError.type === 'adError' &&
+          adError.getErrorCode() === 1009 /*'The VAST response document is empty.'*/ ) {
+          // Empty response happens when 2 or more ads must be loaded simultaneously after playback seek ahead
+          // for 2 or more points of ads. In this case all prev ads loads are cancelled and the only last loaded.
+          // The requests for already cancelled ads seems to return empty VAST. Consider this is not an error
+          // @see https://jira.corp.ooyala.com/browse/PLAYER-3832 discussion
+          return;
```
aeng7: Are we going to add some additional checks in the UI like @jpfranco mentioned in PLAYER-3832? If so, will this part still be needed?

PR #275, opened 2018-09-11
Fix/player 3832

zelibobla:
Implemented debouncing of ads requests to prevent Google IMA 1009 error

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -526,8 +526,7 @@ require("../html5-common/js/utils/utils.js");
       });
 
       /**
-       * debouncing the case when user forwarded ahead and 2 or more ads requests simultaneosly issued
-       * @see https://jira.corp.ooyala.com/browse/PLAYER-3832
+       * Debouncing the case when a user forwarded ahead and 2 or more ad requests are simultaneously issued
        * @private
        * @method GoogleIMA#playAd
        * @param {object} ad The ad to play from the timeline
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -17,6 +17,7 @@ require("../html5-common/js/utils/utils.js");
 
   OO.Ads.manager(function(_, $)
   {
+
     /**
      * @class GoogleIMA
      * @classDesc The GoogleIMA Ads Manager class is registered as an ads manager with the Ad Manager Controller.
@@ -531,6 +532,7 @@ require("../html5-common/js/utils/utils.js");
        */
       this.playAd = function(amcAdPod)
       {
+        
         if(this.currentAMCAdPod)
         {
           _endCurrentAd(true);
@@ -557,7 +559,7 @@ require("../html5-common/js/utils/utils.js");
         {
           IMAiframe.style.zIndex = this.imaIframeZIndex;
         }
-
+        
         if(_usingAdRules && this.currentAMCAdPod.adType == _amc.ADTYPE.UNKNOWN_AD_REQUEST)
         {
           //if the sdk ad request failed when trying to preload, we should end the placeholder ad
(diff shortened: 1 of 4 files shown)


js/google_ima.js
```diff
@@ -524,13 +526,33 @@ require("../html5-common/js/utils/utils.js");
...
+      let debouncerId = null;
       this.playAd = function(amcAdPod)
       {
+        if (debouncerId) {
+          _resetAdsState();
+        }
+        debouncerId = setTimeout(() => {
+          debouncerId = null;
+          playAd(amcAdPod);
+        }, DEBOUNCE_TIMEOUT);
```
aeng7: Pasting my comment from the JIRA ticket:

Is this adding 500 ms to the user experience prior to an ad playback (even a single ad playback), or will this only be used if there are multiple requests in quick succession? If for a single preroll/ad there is now an additional 500 ms of latency, we cannot use this fix/workaround. That adds a significant time to time to first frame for prerolls and when transitioning from content to ads.

js/google_ima.js
```diff
@@ -524,13 +526,33 @@ require("../html5-common/js/utils/utils.js");
       });
 
       /**
-       * Called by the ad manager controller.  Ad Manager Controller lets the module know that an ad should play now.
-       * @public
+       * debouncing the case when user forwarded ahead and 2 or more ads requests simultaneosly issued
```
pilievOoyala: Debouncing the case when a user forwarded ahead and 2 or more ad requests are simultaneously issued

js/google_ima.js
```diff
@@ -524,13 +526,33 @@ require("../html5-common/js/utils/utils.js");
...
+       * @see https://jira.corp.ooyala.com/browse/PLAYER-3832
```
pilievOoyala: Lets keep tickets out of the jsdocs.  External people can't see our tickets. You can add it as a note above the code in the function.

js/google_ima.js
```diff
@@ -524,13 +526,33 @@ require("../html5-common/js/utils/utils.js");
...
+        }, DEBOUNCE_TIMEOUT);
+      }
+
+      /**
+       * Called by the ad manager controller.  Ad Manager Controller lets the module know that an ad should play now.
+       * @private
+       * @method GoogleIMA#playAd
+       * @param {object} ad The ad to play from the timeline.
+       */
+      let playAd = (amcAdPod) => {
```
pilievOoyala: please use the old style of function declaration

KseniiaSokolovaEpam: @pilievOoyala oh, Peter! I am so pity to hear that we have to use old style! Why? 
ES6 is good practice is not it?
... And we use it already in the plugins, skin and mjolnir ...

aeng7: I also agree we should be using ES6. @pilievOoyala what is the reason we should use the old style of function declaration?

pilievOoyala: You guys have convinced me. I see the advantages of the new style and not needing to do a bind.

js/google_ima.js
aeng7: @zelibobla please take a look at this comment. This is a major issue if we are adding 500 ms prior to every ad playback.

zelibobla: Yep, we had to have a call for this, but due to another priority this wasn't discussed. Just forgot then.
At the moment when user clicked forward and got over the ads (and we insist him to see the ads), we yet don't know if he will click immediately again and get over next ads. This is why we need debounce timeout. And yes: we have to use it even for single ads.
It hits UX yes, as I explained originally before implementing this.

js/google_ima.js
zelibobla: Removed

zelibobla:
Closing as outdated

PR #276, opened 2018-09-13
[PLAYER-4414] Avoid use global jQuery reference at vast parser

marcosmx:
Rename global jQuery for Ooyala internal jQuery reference

### DIFF
--- a/js/ooyala_ssai.js
+++ b/js/ooyala_ssai.js
@@ -13,7 +13,7 @@ require("../html5-common/js/utils/constants.js");
 require("../html5-common/js/utils/utils.js");
 require("../html5-common/js/utils/environment.js");
 
-var vastParser = require("../utils/vast_parser.js");
+var VastParser = require("../utils/vast_parser.js");
 var adManagerUtils = require("../utils/ad_manager_utils.js");
 
 OO.Ads.manager(function(_, $)
@@ -42,6 +42,7 @@ OO.Ads.manager(function(_, $)
     this.currentEmbed = "";
     this.domainName = "ssai.ooyala.com";
     this.ssaiGuid = "";
+    this.vastParser = new VastParser($);
 
     this.currentAd = null;
 
@@ -547,7 +548,7 @@ OO.Ads.manager(function(_, $)
     {
       OO.log("Ooyala SSAI: Response");
       // Call VastParser code
-      var vastAds = vastParser.parser(xml);
+      var vastAds = this.vastParser.parser(xml);
       var adIdVastData = _parseVastAdsObject(vastAds);
 
       var adObject = _getAdObjectFromVast(id3Object, adIdVastData);
--- a/utils/vast_parser.js
+++ b/utils/vast_parser.js
@@ -1,4 +1,4 @@
-var VastParser = function() {
+var VastParser = function($) {
 
   this.errorInfo = {};
   this.wrapperParentId = null;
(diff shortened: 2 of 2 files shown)


utils/vast_parser.js
```diff
@@ -123,14 +123,14 @@ var VastParser = function() {
     var nonLinearAds = jqueryXML.find("NonLinearAds");
 
     if (result.type === AD_TYPE.WRAPPER) { result.VASTAdTagURI = jqueryXML.find("VASTAdTagURI").text(); }
-    result.error = filterEmpty(jqueryXML.find("Error").map(function() { return $(this).text(); }));
-    result.impression = filterEmpty(jqueryXML.find("Impression").map(function() { return $(this).text(); }));
-    result.title = _.first(filterEmpty(jqueryXML.find("AdTitle").map(function() { return $(this).text(); })));
+    result.error = filterEmpty(jqueryXML.find("Error").map(function() { return OO.$(this).text(); }));
```
pilievOoyala: I know you aren't adding this line but can you fix the spacing fo this  to agree with our coding guidelines? The return shouldn't be in the same line as the rest of the code, so we can put breakpoints in there to debug if necessary.

Something more like this:
result.error = filterEmpty(jqueryXML.find("Error").map(function() { 
  return OO.$(this).text(); 
}));

And do that for the other two lines below this.

utils/vast_parser.js
```diff
@@ -36,7 +36,7 @@ var VastParser = function() {
    * @returns {object[]} An array containing the ad(s) if ads are found, otherwise it returns null.
    */
   this.parser = function(vastXML, adLoaded) {
-    var jqueryAds =  $(vastXML).find("Ad");
+    var jqueryAds =  OO.$(vastXML).find("Ad");
```
pilievOoyala: I'm confused as to why we need this. Is $ not defined anymore in this scope?  This may be a bigger issue.

jpfranco: So this is because $ used to be available in the global scope due to a bug we had. The main plugin code does have a reference to jquery that is passed by the AMC, but this `VastParser` function doesn't have it scope.

I think it's just a matter of doing something:
`var VastParser = function($) {`

And then just pass the reference from the AMC when initializing. This way we avoid modifying all occurrences of $

marcosmx: Yes, the '$' it's not available in the scope of the library/util, it seems that it was working before because it was defined at the global scope (window) and it was not detected using the debug page because that web app exposes the jQuery symbol globally...I found this trying to build a custom player.

utils/vast_parser.js
marcosmx: done

utils/vast_parser.js
jpfranco: @marcosmx could we just pass the reference with a parameter like on the example above? Even though using `OO.$` works, I think it might be better to avoid globals and it might also be a bit cleaner if we just keep using $

marcosmx: sure, let me do the proper changes and I'll let you know once it's done

marcosmx: @jpfranco @pilievOoyala please check the new approach

js/ooyala_ssai.js
```diff
@@ -42,6 +42,7 @@ OO.Ads.manager(function(_, $)
     this.currentEmbed = "";
     this.domainName = "ssai.ooyala.com";
     this.ssaiGuid = "";
+    this.vastParser = new VastParser($)
```
jpfranco: Missing semicolon.

marcosmx: done

jpfranco:
@marcosmx btw, the SSAI plugin is the only one that uses VastParser correct?

marcosmx:
> @marcosmx btw, the SSAI plugin is the only one that uses VastParser correct?

Yeah, I did a quick search in all the project and just SSAI uses this util

jpfranco:
@pilievOoyala can you check the latest updates? I think SSAI will currently not work unless there's jQuery in the page until we merge these changes.

aeng7:
LGTM

PR #279, opened 2018-11-02
Commented out failed tests

zelibobla:


aeng7:
@zelibobla @DmitriiMaslov will these tests be fixed in a future commit? We should not be removing unit tests if the scenario they test still applies.

paparajupavan:
This is not ideal to just comment the unit tests  and pushing the fixes.Is there any reason why we did this? @DmitriiMaslov @zelibobla

zelibobla:
The reason is to unblock testing team. 2nd of Nov is the date when candidate had to be issued (and it had).
I've asked @dallen42 (author of the failing tests) to fix the tests and set up a PR.

PR #282, opened 2018-11-12
Player 1340 - Enable preloading of IMA ads

aeng7:
I'll add unit tests in the near future. This is to get QA started on testing these changes.

### DIFF
--- a/js/google_ima.js
+++ b/js/google_ima.js
@@ -44,7 +44,7 @@ require("../html5-common/js/utils/utils.js");
       var _usingAdRules;
       var _IMAAdsLoader;
       var _IMAAdsManager;
-      var _IMAAdsManagerInitialized;
+      var _imaAdPlayed;
       var _IMAAdDisplayContainer;
       var _linearAdIsPlaying;
       var _timeUpdater = null;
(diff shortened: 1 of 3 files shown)


js/google_ima.js
```diff
@@ -528,9 +530,39 @@ require("../html5-common/js/utils/utils.js");
...
+      this.playAd = function(amcAdPod, adRequestOnly)
       {
+        if (amcAdPod === null || typeof amcAdPod === 'undefined')
+        {
+          return;
+        }
+
+        if (!adRequestOnly)
+        {
+          _playAdOnRequestSuccess = amcAdPod;
```
jpfranco: This sounds a bit like a boolean. Maybe we could rename it to something like `_adToPlayOnRequestSuccess`

test/unit-test-helpers/mock_ima.js
```diff
@@ -146,10 +146,6 @@ google =
                 amCallbacks[event] = callback;
               };
               this.start = function() {
-                //Use the canPublishEvents flag to ensure that start
-                //gets called, otherwise publish events will fail,
-                //failing unit tests
-                canPublishEvents = true;
```
aeng7: We no longer need this since we no longer have setAutoPlayAdBreaks set to false.

PR #287, opened 2019-01-15
[PLAYER-5008] set current offset before start

ooyala-amohnatkin:
SSAI backend requires current offset from the end. When video starts we should already know video duration to set it as offset.
`CONTENT_TREE_FETCHED` fires before start

### DIFF
--- a/js/ooyala_ssai.js
+++ b/js/ooyala_ssai.js
@@ -162,6 +162,11 @@ OO.Ads.manager(function(_, $)
       amc.addPlayerListener(amc.EVENTS.AD_VOLUME_CHANGED, _.bind(this.onAdVolumeChanged, this));
       amc.addPlayerListener(amc.EVENTS.MUTE_STATE_CHANGED, _.bind(this.onMuteStateChanged, this));
       amc.addPlayerListener(amc.EVENTS.PLAY_STARTED, _.bind(this.onPlayStarted, this));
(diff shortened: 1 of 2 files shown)


js/ooyala_ssai.js
```diff
@@ -162,6 +162,11 @@ OO.Ads.manager(function(_, $)
       amc.addPlayerListener(amc.EVENTS.AD_VOLUME_CHANGED, _.bind(this.onAdVolumeChanged, this));
       amc.addPlayerListener(amc.EVENTS.MUTE_STATE_CHANGED, _.bind(this.onMuteStateChanged, this));
       amc.addPlayerListener(amc.EVENTS.PLAY_STARTED, _.bind(this.onPlayStarted, this));
+      amc.addPlayerListener(amc.EVENTS.CONTENT_TREE_FETCHED, _.bind(this.onContentTreeFetched, this));
+    };
+
+    this.onContentTreeFetched = function (event, content) {
+      currentOffset = content.duration;
```
KseniiaSokolovaEpam: I believe we can add tests here?

ooyala-amohnatkin: sure 🤥

ooyala-amohnatkin: but it's a blocker
tests are coming in the next pr

PR #295, opened 2019-03-19
Fixing the adtimer for different use cases.

paparajupavan:
1.Pausing Ad
2.For ad events which are not generated for different ad percentages


js/ooyala_ssai.js
```diff
@@ -366,11 +368,21 @@ OO.Ads.manager(function()
     this.resumeAd = (ad) => {
       if (adMode) {
         _handleTrackingUrls(this.currentAd, ["resume"]);
-        if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id] && isFinite(ad.duration)) {
+        if (ad && ad.ad && ad.ad.data && this.adIdDictionary[ad.ad.data.id] && _.isFinite(ad.duration)) {
+          var duration = ad.duration;
+          //Deducting the already played duration of ad  from the actual ad duration for making timer accurate
+          if(_.isFinite(this.adIdDictionary[ad.ad.data.id].startTime) && _.isFinite(this.adIdDictionary[ad.ad.data.id].pauseTime) )
```
ooyala-amohnatkin: it's better to use destructuring to increase code readability
`const { startTime, pauseTime } = this.adIdDictionary[ad.ad.data.id];`

js/ooyala_ssai.js
```diff
@@ -519,15 +534,22 @@ OO.Ads.manager(function()
...
-     * @method OoyalaSsai#onReplay
-     */
-    this.onReplay = () => {
-      currentOffset = 0;
-      this.currentAd = null;
-    };
+    *
+    *
+    */
+    var _getAdDuration = _.bind(function(id3Object)
```
ooyala-amohnatkin: Please use an arrow function and remove `bind`
```
var _getAdDuration = (id3Object) => {
}
```

js/ooyala_ssai.js
paparajupavan: Incorporated the changes.

js/ooyala_ssai.js
paparajupavan: Incorporated the changes.

paparajupavan:
> Please check that tests are passing.
> There's no `_` variable, all underscore methods are imported directly, so `_.isFinite` should be `isFinite`
> The same for `delay`

Yes.Changed the functions.

paparajupavan:
> Please check that tests are passing.
> There's no `_` variable, all underscore methods are imported directly, so `_.isFinite` should be `isFinite`
> The same for `delay`

Modified it.

PR #300, opened 2019-04-04
[PLAYER-5418] - Not able to play VR360 videos on IOS

ipadilla4:
The issue was that ads were not displayed because DOM video element was not found in Safari and Chrome. Solution consist in modifying css selector to make it compatible with Safari.


ipadilla4:
Closing this ticket, I made an error in the ticket number.

PR #305, opened 2019-04-15
feature/PLAYER-5346

MariskaS:
migrate from Gulp to WebPack


index.js
```diff
@@ -0,0 +1,12 @@
+const OO = {};
```
alexeyMohnatkin: you override `OO` object
please, describe what is this file for

webpack.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = () => ({
   optimization: {
     minimize: true,
   },
+  mode: 'development',
```
ooyala-amohnatkin: `mode` is set in CLI options, no need to set it here

ooyala-amohnatkin:
Close because of the wrong ticket number.

